PR #3, opened 2019-10-05
Remove old folder and updated script file to test the webhook

faucetcollector:


zec faucet.info/ZEC Faucet.info.fc
```diff
@@ -1,6 +1,6 @@
 using System;
 using FaucetCollector.Script;
-//edit
+//edit, test from github
```
faucetcollector: You can leave comments here

PR #6, opened 2019-10-06
Update ClaimBtc.in.fc

aviramdayan:
update timer

### DIFF
--- a/ClaimBtc.in/ClaimBtc.in.fc
+++ b/ClaimBtc.in/ClaimBtc.in.fc
@@ -2,7 +2,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 27/25/2019 21:09PM
-//Last Update: 26/08/2019 19:20PM - Version 4
+//Last Update: 06/10/2019 19:20PM - Version 10
 //==================//
 using System;
 using System.Collections.Generic;
@@ -58,9 +58,19 @@ public class ClaimBtcInScript_ : FaucetScript
         //return base.IsLoggedIn();
     }
 
+
+//=============//
+//GetFaucetWaitTime
+//=============//
     public override int GetFaucetWaitTime ( )
     {
         var WaitTimerText = ElementByXPath("//span[@class='kkcountdown-box']");
+
+        if (!IsVisible(WaitTimerText))
+        {
+		  return GetWaitSetting ( );
+        }	
+
         if ( IsVisible ( WaitTimerText ) )
         {
             var minVAR = ElementByXPath("//span[@class='kkc-min']");
@@ -76,6 +86,9 @@ public class ClaimBtcInScript_ : FaucetScript
 
         return base.GetFaucetWaitTime ( );  
     }
+//=============//
+//GetFaucetWaitTime
+//=============//
 
 
     //========//
--- a/ClaimBtc.in/ClaimBtc.in.fc
+++ b/ClaimBtc.in/ClaimBtc.in.fc
@@ -1,4 +1,3 @@
-﻿
 
 //==================//
 //script Written by: ••AviramDayan•• 27/25/2019 21:09PM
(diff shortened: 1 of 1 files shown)


ClaimBtc.in/ClaimBtc.in.fc
```diff
@@ -48,6 +47,258 @@ public class ClaimBtcInScript : FaucetScript
...
+			
+             Log("FaucetClaim: WaitTimerText apper"); 
+             int min,sec;
+             min = Convert.ToInt32(minVAR.Text)*60;
+             sec = Convert.ToInt32(secVAR.Text);
+             return min+sec;
+
+        } // end WaitTimerText
+		
+
```
FCCristian: Problem with this is that timer get at the last of claim. if for example the last claim fail the script will not return any time. In the last of the cases we need to get time of the first claim.

ClaimBtc.in/ClaimBtc.in.fc
```diff
@@ -48,6 +47,258 @@ public class ClaimBtcInScript : FaucetScript
...
     public override bool IsLoggedIn()
+
+//==================//
+//script Written by: ••AviramDayan•• 27/25/2019 21:09PM
+//Last Update: 26/08/2019 19:20PM - Version 4
+//==================//
+using System;
+using FaucetCollector.Script;
+//css_inc IconCaptcha
+public class ClaimBtcInScript : FaucetScript
```
FCCristian: You have added this two times in the script

ClaimBtc.in/ClaimBtc.in.fc
aviramdayan: so also this is good
        if (!IsVisible(WaitTimerText))
        {
		  return GetWaitSetting ( );
	}	


`
//=============//
//GetFaucetWaitTime
//=============//
 public override int GetFaucetWaitTime()
 {

        //time
        var WaitTimerText = ElementByXPath("//span[@class='kkcountdown-box']");
		
        if (!IsVisible(WaitTimerText))
        {
		  return GetWaitSetting ( );
	}		
		
        if (IsVisible(WaitTimerText))
        {
             var minVAR = ElementByXPath("//span[@class='kkc-min']");
             var secVAR = ElementByXPath("//span[@class='kkc-sec']");
			
             Log("FaucetClaim: WaitTimerText apper"); 
             int min,sec;
             min = Convert.ToInt32(minVAR.Text)*60;
             sec = Convert.ToInt32(secVAR.Text);
             return min+sec;

        } // end WaitTimerText
		
	
   //let Faucet Collector continue.
   return base.GetFaucetWaitTime();
 }
//=============//
//GetFaucetWaitTime
//=============//`

ClaimBtc.in/ClaimBtc.in.fc
```diff
@@ -48,6 +47,258 @@ public class ClaimBtcInScript : FaucetScript
...
+            Log ( Success.Text );
+            return 0;
+        }
+        else
+        {
+            var Fail_=ElementByXPath(FailXPath);
+            if ( IsVisible ( Fail_ ) )
+            {
+                Log ( Fail_.Text );
+            }
```
faucetcollector: Why did you write custom code, and not do the default FC code that checks for the success/fail?

aviramdayan: > Why did you write custom code, and not do the default FC code that checks for the success/fail?

the question is directed at christian  @FCCristian

FCCristian: The reason was that the method is called every time a faucet is claimed, the user can select a lot of faucets to claim. so if for example he select the 12 faucets he will see 12 success instead of one. So like that we can get the success and mark only one claim success if at least one of the 12 faucets selected was claimed.

ClaimBtc.in/ClaimBtc.in.fc
FCCristian: it's not good because if faucet claim fail this element will not exist '//span[@class='kkcountdown-box']'

ClaimBtc.in/ClaimBtc.in.fc
```diff
@@ -58,9 +58,19 @@ public class ClaimBtcInScript_ : FaucetScript
...
+//=============//
     public override int GetFaucetWaitTime ( )
     {
         var WaitTimerText = ElementByXPath("//span[@class='kkcountdown-box']");
+
+        if (!IsVisible(WaitTimerText))
+        {
+		  return GetWaitSetting ( );
+        }	
+
```
FCCristian: This will not work. The script will check this every time. Even if the faucet is not claimed yet. So will return time. In my previus update was worked well,

aviramdayan: its not true, we call it in the end, it work

aviramdayan: Maybe Confirm it already, it's better than what exists before

aviramdayan: approve this script and better to concentrate mainly on solving "solve icon"

aviramdayan: I don't have the power to get stuck all the time on small things, 
have to run scripts..
You have to concentrate mainly and not small
If that will be the case all the time, I raise my hands.

FCCristian: Mate... , you are doing it on GetFaucetWaitTime. The script will return wait 100% of times this makes no sense.

aviramdayan: image
here is a proof from now

ClaimBtc.in/ClaimBtc.in.fc
```diff
@@ -383,11 +152,17 @@ public class ClaimBtcInScript : FaucetScript
 //==============//
 //DoSolveFaucet
 //==============//
- public override int DoSolveFaucet()
-  {
 
+private class Timer
+    {
+        public int Seconds { get; set; }
+        public DateTime Date { get; set; } = DateTime.Now;
```
FCCristian: @faucetcollector  this throw error of syntax inside of faucetcollector but inside of VS the syntax look correct, Do you know why?  What version of c # is the compiler using?

ClaimBtc.in/ClaimBtc.in.fc
FCCristian: it is impossible that the photo is from this version of the script since it has a syntax error and you cannot compile inside of FC.

ClaimBtc.in/ClaimBtc.in.fc
faucetcollector: Yes, you can only use C# 5 syntax.
So auto-implemented properties don't work as that is a C# 6 feature.

PR #12, opened 2019-10-09
Update Free-Ethereum.fc

FCCristian:
->Fixed withdrawal for the script
->Fixed error callback not found
->Delete unnecessary using statements
->clear the code.


JReming85:
question, do you think each "fix/update" should be its own patch/revision and bundle them all in the same pr, or just do them al in a single fix?

faucetcollector:
Its fine if you make multiple changes in the same PR - but for processing it might be more easy to do a PR per script instead of a single larger one

PR #15, opened 2019-10-11
Update FreeBitcoin.io.fc

cgerabit:
*Delete innesesary settings, The only functionality of the script is claim faucet  so we don't need that setting
*Fixed Withdrawal.
*deleted "//" from the url
*Moved shortlink to AfterSolveFaucet
*Moved withdrawal to BeforeSolveFaucet

### DIFF
--- a/FreeBitcoin.io BTC/FreeBitcoin.io.fc
+++ b/FreeBitcoin.io BTC/FreeBitcoin.io.fc
@@ -16,131 +16,47 @@ using System.Globalization;
 
 public class FreebitcoinioScript : FaucetScript
 {
-    /// <summary>
-    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
-    /// You can get the value the user entered with the methods: GetSetting("[Name of the FaucetSetting]"), GetBoolSetting and GetDateTimeSetting
-    /// You can also create a new setting value using SetSetting("[Name you want to use]", "value")
-    /// </summary>
     public override FaucetSettings Settings
     {
         get
         {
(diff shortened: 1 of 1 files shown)


FreeBitcoin.io BTC/FreeBitcoin.io.fc
```diff
@@ -160,77 +76,33 @@ public class FreebitcoinioScript : FaucetScript
         SetText(signInEmail, GetSetting("Email"));
     
 		SetText(signInPassword, GetPassword("Password"));
- 
-		
+
+
         //Now we search for the login submit button
-        var loginButton = ElementByClass("login");
+        var loginButton = ElementByXPath("//button[@class='main-button main-button-yellow login bg-3']");
```
faucetcollector: quite some classes for a login button, might wanna do a [contains(@class, 'login')] - if there are no others using that class also, otherwise add another class to the search [contains(..) and contains(..)]

FreeBitcoin.io BTC/FreeBitcoin.io.fc
```diff
@@ -160,77 +76,33 @@ public class FreebitcoinioScript : FaucetScript
...
-        {
-            Click(loginButton);
-        }
+         Click ( loginButton );
 
         Wait();
         var Error = ElementByXPath("//div[@class='error']");
         if (IsVisible(Error))
         {
+            if ( Error.Text.Contains ( "email" ) || Error.Text.Contains("password") )
```
faucetcollector: Just a note: Contains is case sensitive

PR #16, opened 2019-10-11
 BonusBitcoin.co

aviramdayan:
fix-login-drive

PR #17, opened 2019-10-11
Bitfun.co

aviramdayan:
fix-login-drive

### DIFF
--- a/Bitfun.co/Bitfun.co.fc
+++ b/Bitfun.co/Bitfun.co.fc
@@ -48,6 +48,475 @@ public class BitfunScript : FaucetScript
     public int SolveCapt()
     {
         var XPathAdcopyRefresh = "//a[@id='adcopy-link-refresh']";
+
+using System;
+using System.Text.RegularExpressions;
+using FaucetCollector.Script;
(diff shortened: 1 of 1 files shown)


Bitfun.co/Bitfun.co.fc
```diff
@@ -48,6 +48,475 @@ public class BitfunScript : FaucetScript
     public int SolveCapt()
     {
         var XPathAdcopyRefresh = "//a[@id='adcopy-link-refresh']";
+
+using System;
```
faucetcollector: Looks like the file isnt correct, it starts here with a new part

aviramdayan: this part i don't change (it was before)
`var XPathAdcopyRefresh = "//a[@id='adcopy-link-refresh']";`
image

this answer probably related to "BonusBitcoin.co" also
https://github.com/Faucet-Collector/scripts/pull/16

faucetcollector: Right now it looks like you pushed a scripts inside a script, can you check what is on your local disk? 
Same for PR #16 yes.

PR #19, opened 2019-10-14
Cryptoo.Me

aviramdayan:
Protection-and-waiting-time-changes

### DIFF
--- a/Cryptoo.Me/Cryptoo.Me.fc
+++ b/Cryptoo.Me/Cryptoo.Me.fc
@@ -1,8 +1,8 @@
-﻿
+
 
 //==================//
 //script Written by: ••AviramDayan•• 04/05/2019 10:30AM
-//Last Update: 27/09/2019 10:30AM - Version 5
+//Last Update: 14/10/2019 10:30AM - Version 8
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


Cryptoo.Me/Cryptoo.Me.fc
```diff
@@ -379,7 +377,17 @@ public class cryptoome : FaucetScript
...
 	    Wait(15);
-	
+		
+        var Error_NeedToWait = ElementByXPath("//div[@class='alert alert-danger' and contains(text(),'wait') ]"); 
+        if (IsVisible(Error_NeedToWait))
+		{
+		Log("FaucetClaim - Error_NeedToWait ");  
+        var WaitVAR = ElementByXPath("//div[@class='alert alert-danger'] | //p[@class='alert alert-info']");
+        int min;
+        min = Convert.ToInt32(WaitVAR.Text.Split(' ')[4])*60;
```
faucetcollector: Assuming there are 5 items in the array from the `Split` is generally not a good idea, better move this logic to a separate method (as the same code is used multiple times) and check the length of the array that is returned by `Split`. 
It is also recommended to use `int.TryParse`.

aviramdayan: > Assuming there are 5 items in the array from the `Split` is generally not a good idea, better move this logic to a separate method (as the same code is used multiple times) and check the length of the array that is returned by `Split`.
> It is also recommended to use `int.TryParse`.

It is possible to move to one void function (I will do now)
But in general, I left it that way
because it could be the element of "minutes" appears before the iframe
(as it was on the site before) It can come back so I left that double
The "Convert.ToInt32" works - not critical ..

PR #21, opened 2019-10-18
Update FreeBitco.in.fc

cgerabit:
Solved the problem explained in the ticket #0239

### DIFF
--- a/FreeBitco.in/FreeBitco.in.fc
+++ b/FreeBitco.in/FreeBitco.in.fc
@@ -243,14 +243,17 @@ public class FreeBitcoinScript_ : FaucetScript
             
             for(int i = 0; i < DoubleCaptchaImg.Count; i++)
             {
+                int aux = i;
                 TaskList.Add(Task.Run(() =>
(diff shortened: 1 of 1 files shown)


FreeBitco.in/FreeBitco.in.fc
```diff
@@ -243,14 +243,17 @@ public class FreeBitcoinScript_ : FaucetScript
...
             {
+                int aux = i;
                 TaskList.Add(Task.Run(() =>
                 {
-                    return DoSolveImageCaptcha(DoubleCaptchaImg[i], DoubleCaptchaBox[i]);
+                        return DoSolveImageCaptcha(DoubleCaptchaImg[aux], DoubleCaptchaBox[aux]);                  
                 })
                 );
+                Wait(1);
             }
```
cgerabit: In some occasions and within the task 'i' was take value of the end of the iteration and this threw an exception of type out of range.
this problem seems to be solved using aux int var and also adding wait for every task

PR #22, opened 2019-10-19
Feature/Mooncash.io/Add current balance to log output

JReming85:
Adds a info entry to show the accounts total balance


MoonCash.co/MoonCash.co.fc
```diff
@@ -503,6 +506,13 @@ public class MoonCashScript : FaucetScript
     {
         return base.CheckFaucetResult();
         //This faucet has nothing to do here. We will let Faucet Collector handle it.
+	/// Outputs Current Balance in log
+	var currentBalance = ElementByXPath("//a[contains(@data-bind,'text: balance().toFixed(8)')]");
```
faucetcollector: the balance element its data-bind attribute contains this? its different from your other commits. can't check atm because the coinpot faucets are down

JReming85: yea, cash and dash both use a different model then the rest of the moonsites.

The only downside to these sites is they abrev the balance (I think) 100k, 200k, etc but will still give a good idea on when its time to withdraw

faucetcollector:
This PR also contains (old) changes for MoonLiteco.in.fc, which are already merged in #30

PR #24, opened 2019-10-19
Freature/moon bitcoin/add balance to log

JReming85:
Add info entry in log showing current balance


MoonBitco.in/MoonBitco.in.fc
```diff
@@ -343,6 +343,15 @@ public class MoonBitcoinScript : FaucetScript
...
         }
+		
+		var currentBalance = ElementByXPath("//span[contains(@title,'balance:')]");
+		if (IsVisible(currentBalance));
+			{
+				Log("Balance: " + currentBalance.Text + " BTC");
+				Wait(1);
+				ClaimSuccess();
+			}
+
```
cgerabit: Put ClaimSuccess() in BeforeSolveFaucet is not a good idea, because even if the claim fails it will be marked satisfactory
one solution to put the balance after claim the faucet is doing it in CheckFaucetResult()

You can try something like this:

 public override int CheckFaucetResult()
    {
        var result= base.CheckFaucetResult();

        var currentBalance = ElementByXPath("//span[contains(@title,'balance:')]");
        if (IsVisible(currentBalance)) 
        {
            Log("Balance: " + currentBalance.Text + " BTC");

        }
        return result;
    }

MoonBitco.in/MoonBitco.in.fc
```diff
@@ -343,6 +343,15 @@ public class MoonBitcoinScript : FaucetScript
             Log("Seems that the site is offline, Cloudflare is visible you need to wait");
             return GetFaucetWaitTime();
         }
+		
+		var currentBalance = ElementByXPath("//span[contains(@title,'balance:')]");
+		if (IsVisible(currentBalance));
```
cgerabit: Take care with the ';' the if sentence does not wear it

JReming85: thanks, completely missed that one

MoonBitco.in/MoonBitco.in.fc
JReming85: awesome thanks, I think I was just so glad it worked I didnt care where it was (noticed it was adding +1 to the claim afterwards) so this makes a lot of sense.

JReming85:
updated, chechBalance is now at the end in CheckFaucetResult()

MoonBitco.in/MoonBitco.in.fc
```diff
@@ -467,14 +467,12 @@ public class MoonBitcoinScript : FaucetScript
...
-		}
-		return result;
-	
+        return base.CheckFaucetResult();
+        var currentBalance = ElementByXPath("//span[contains(@title,'balance:')]");
+        if (IsVisible(currentBalance));
+          {
+            Log("Balance: " + currentBalance.Text + " BTC");
+          }
+          return result;
```
cgerabit: you cannot return base.CheckFaucetResult() and them put the code to see the balance because the code after the return will not be executed, even if you try to run this script you will see compilation error like this: 

> the name result does not exist in the current context

It's because you are returning result but result is not declared. if you see my past example i'm asigning the value of base.CheckFaucetResult() to a variable. and I'm returning it after log the balance

base.CheckFaucetResult() is the default implementation of CheckFaucetResult, in the finally this is a like a normal function and  will return int value , so in my past example we are asign that value to return it after log the balance

btw and remember delete the ';' after the if otherwhise the if will not work and the Log will be executed 100% of times , even if the element to see the balance is not visible. If this happen it will return a exception

MoonBitco.in/MoonBitco.in.fc
```diff
@@ -465,6 +467,14 @@ public class MoonBitcoinScript : FaucetScript
     }
     public override int CheckFaucetResult()
     {
-        return base.CheckFaucetResult();
+        var result= base.CheckFaucetResult();
+		
+		var currentBalance = ElementByXPath("//span[contains(@title,'balance:')]")
```
faucetcollector: missing `;` on the end

PR #27, opened 2019-10-19
Feature/moonltc/add balance to log

JReming85:
adds balance to log


faucetcollector:
moonlite is also included in  https://github.com/Faucet-Collector/scripts/pull/22

faucetcollector:
we can use the update for moonlite in #22, will close this one.

PR #29, opened 2019-10-20
Update adBTC.fc

cgerabit:
*Improved some the code.
*Fixed surf ads if the language of the site is not the default ( english)

### DIFF
--- a/adBTC/adBTC.fc
+++ b/adBTC/adBTC.fc
@@ -9,19 +9,15 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 //css_ref DoubleMathCaptchaUtility.dll
 namespace Adbtc
 {
-
-
     public class TemplateScript : FaucetScript
     {
-        /// <summary>
-        /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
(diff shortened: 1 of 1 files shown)


adBTC/adBTC.fc
```diff
@@ -31,149 +27,95 @@ namespace Adbtc
...
-            if (IsVisible(log))
-            {
-                Click(log);
-            }
-
-            bool Bypassed = true;
-            Wait(4);
 
-            if (IsVisible(ElementByXPath("//img[contains(@src,'adbtc.top/1captcha')]")))
+            IWebElement AddressBox = ElementByXPath("//input[@id='addr']");
```
faucetcollector: preferably use `var addressBox = ElementByXPath..`

adBTC/adBTC.fc
```diff
@@ -240,501 +185,151 @@ namespace Adbtc
...
-                        Click(drop);
-                        Wait(2);
-                        int result = SolveCaptcha();
-                        if (result > 0)
-                        {
-                            return Fail("FAil solve Captcha");
-                        }
+                        Driver.Navigate().Refresh();
+                        Wait(1);
+                        continue;
```
faucetcollector: Maybe it will be good to limit this also with a max attempts

adBTC/adBTC.fc
cgerabit: Yes , the code cleaner changed them and I didn't realize

PR #35, opened 2019-10-23
Update Free-Litecoin.com.fc

cgerabit:
*Changed password box from EditorType.Wallet to EditorType.Password


Free-Litecoin.com/Free-Litecoin.com.fc
```diff
@@ -15,7 +15,7 @@ public class freeLitecoinScript : FaucetScript
             return new FaucetSettings("https://www.free-litecoin.com/")
             {
                 new FaucetSetting() { Name = "Email", Display = "Email Address", Type = EditorType.TextBox, Required = true },
-                new FaucetSetting() { Name = "ps", Display = "password", Type = EditorType.Wallet, Required = true },
+                new FaucetSetting() { Name = "ps", Display = "password", Type = EditorType.Password, Required = true },
```
faucetcollector: Can you introduce a new name? For existing customers there is something in the ps setting already - the password field is encrypted, but existing values wont be. So best to use a new name - that will make the users need to enter it first because its required

PR #44, opened 2019-10-30
Update FreeBitco.in.fc

cgerabit:
*Fixed required points settings


FreeBitco.in/FreeBitco.in.fc
```diff
@@ -517,13 +517,33 @@ public class FreeBitcoinScript_ : FaucetScript
...
+        int Required;
+        if(int.TryParse(GetSetting(SettingName),out Required ) )
+        {
+            return Required;
+        }
+        else
+        {
+            Log ( "There is a problem reading setting of points required for "+SettingName );
+        }
+        return 0;
```
faucetcollector: better return a very high value - otherwise it will try to redeem this option

PR #47, opened 2019-11-02
Feature/voxin adz/first release

cgerabit:


VoxinAdz/VoxinAdz.fc
```diff
@@ -0,0 +1,301 @@
...
+                return Fail
+                    ( "Button login is not visible" );
+            }
+            Click
+                ( ButtonLogin );
+
+            Wait ( );
+
+            var ErrorLogin=ElementByXPath(FailLoginXPath);
+            if ( !IsVisible ( ErrorLogin ) )
```
faucetcollector: Should be `IsVisible` instead of `!IsInvisible`

PR #48, opened 2019-11-02
Update VoxinAdz.fc

cgerabit:
Fixed compilation errors from the past version


cgerabit:
Sorry for the mistake in the past version , i've disabled script meanwhile the fix is aproved.

PR #49, opened 2019-11-04
Create WortexFaucet.fc

cgerabit:
First Commit

### DIFF
--- a/Wortex-Faucet/WortexFaucet.fc
+++ b/Wortex-Faucet/WortexFaucet.fc
@@ -120,9 +120,15 @@ namespace ScriptsOfFc
             var Timer=ElementByXPath("//div[@id='t']");
             if ( IsVisible ( Timer ) )
             {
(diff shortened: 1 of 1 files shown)


Wortex-Faucet/WortexFaucet.fc
```diff
@@ -0,0 +1,130 @@
...
+            Wait ( );
+            
+            return base.DoSolveFaucet ( );
+        }
+        public override int GetFaucetWaitTime ( )
+        {
+            var Timer=ElementByXPath("//div[@id='t']");
+            if ( IsVisible ( Timer ) )
+            {
+                var TimerText=Timer.Text.Replace("min","").Replace("seg","").Trim().Split(new[]{ ' ' },StringSplitOptions.RemoveEmptyEntries);
```
faucetcollector: Is it seg or sec?
And best to check the number of items in TimerText and use int.TryParse

PR #51, opened 2019-11-04
Bloxawards.com

aviramdayan:
new script

### DIFF
--- a/Bloxawards.com.fc
+++ b/Bloxawards.com.fc
@@ -1,0 +1,224 @@
+//==================//
+//script Written by: ••AviramDayan•• 05/11/2019 00:00PM - Version 1
+//Last Update: 05/11/2019 00:00PM - Version 1
+//==================//
+
+using System;
+using FaucetCollector.Script;
+
+public class BloxawardsScript : FaucetScript
+{
+
+    public override FaucetSettings Settings
+    {
+        get
+        {
+            return new FaucetSettings(DomainName_String)
+            {
+                new FaucetSetting() { Name = "Email", Display = "E-mail", Type = EditorType.TextBox, Required = true },
+                new FaucetSetting() { Name = "Password", Display = "Password", Type = EditorType.Password, Required = true },
+            };
+        }
+    }
+
+    string DomainName_String = "https://bloxawards.com/";
+
+    public override void Start()
+    {
+        //Title that shows in the browser. Is used to identify and close popup windows
+        Title = "Bloxawards.com";
+
+        //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
+        SuccessXPath = "//*";
+        FailXPath = "//*";
+
+        //Let Faucet Collector start up everything
+        base.Start();
+    }
+
+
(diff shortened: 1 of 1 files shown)


Bloxawards.com.fc
```diff
@@ -0,0 +1,224 @@
...
+        //utility.DoSolve();
+        if (result > 0)
+        {
+          return Fail("fail not solve"); //return to solve again - if something wrong
+        }
+
+        Wait();	
+	    Wait(2);
+		
+        var ButtonLog = ElementByXPath("/html/body/div/div[4]/div/div/form/button");
```
faucetcollector: Best would be to find a proper small xpath for this element

Bloxawards.com.fc
```diff
@@ -0,0 +1,224 @@
...
+
+    string DomainName_String = "https://bloxawards.com/";
+
+    public override void Start()
+    {
+        //Title that shows in the browser. Is used to identify and close popup windows
+        Title = "Bloxawards.com";
+
+        //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
+        SuccessXPath = "//*";
```
faucetcollector: No success/fail xpath?

Bloxawards.com.fc
```diff
@@ -0,0 +1,224 @@
...
+
+
+
+
+    public override int DoSolveFaucet()
+    {
+    	
+        GoToUrl(DomainName_String+"/giveaway");
+        Wait();		
+        Wait(2);
```
faucetcollector: These two lines are the same as Wait(3);
or Wait(2, 3);

Bloxawards.com.fc
aviramdayan: yeah for sure not i write this section, i will fix it

Bloxawards.com.fc
aviramdayan: tahre is no success/fail xpath
so i did in the DoSolveFaucet protection

`
    public override int DoSolveFaucet()
    {
    	
        GoToUrl(DomainName_String+"/giveaway");
        Wait();		
        Wait(2);		
	
        var claimButton = ElementByClass("btn-main2");
		
        if (!IsVisible(claimButton))
        {
            //We couldn't find the button, return we have a failure.
            Log("Claim button not found or not visible.");
	    return GetWaitSetting(); //fail
        }
        
        var resultClaim = SolveCaptcha();
        if (resultClaim > 0)
        {
            //Unable to solve the captcha - return the failure
            return resultClaim;
        }
		
        if (IsVisible(claimButton))
        {
	    Log("claimButton - now");
            Click(claimButton);
        }	
		
		Wait(3);
		
	ad = true;    // success
        //return GetWaitSetting();	
		
        return base.DoSolveFaucet();
    }
`

Bloxawards.com.fc
aviramdayan: Before he goes to the site, I always put in first Wait (); (Clean)
After it loads i give another Wait (2); for protection I found this very useful

PR #55, opened 2019-11-06
FreeFaucet.io.fc

aviramdayan:
new_created

### DIFF
--- a/FreeFaucet.io/FreeFaucet.io.fc
+++ b/FreeFaucet.io/FreeFaucet.io.fc
@@ -7,6 +7,10 @@ using System;
 using System.IO;
 using FaucetCollector.Script;
 using OpenQA.Selenium;
+
+using OpenQA.Selenium.Support.UI;
+using System.Linq;
+
 //css_inc ShortLinkUtilities
 //css_inc RecaptchaUtilities
 
@@ -98,12 +102,12 @@ public class BloxawardsScript : FaucetScript
         Wait();	
 	    Wait(2);		
 	
-        var box_user = ElementByXPath("//div[2]/div[2]/form/div[1]/input[@name='user_name']");
+        var box_user = FindVisibleElement_Function(By.XPath("//input[@id='login_input_email']")); 
         SetText(box_user, GetSetting("Email"));
 		
         Wait(2);
 		
-        var box_pass = ElementByXPath("//div[2]/div[2]/form/div[2]/input[@name='user_password']");
+        var box_pass = FindVisibleElement_Function(By.XPath("//input[@id='login_input_password']"));
         SetText(box_pass, GetPassword("Password"));
 		
         Wait(2);
@@ -120,7 +124,7 @@ public class BloxawardsScript : FaucetScript
         Wait();	
 	    Wait(2);
 		
-        var ButtonLog = ElementByXPath("//div[2]/div[2]/form/input[@type='submit']");
+        var ButtonLog = FindVisibleElement_Function(By.XPath("//input[@type='submit']")); 
         Click(ButtonLog);
 
         Wait(2);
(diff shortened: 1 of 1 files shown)


FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+            return new FaucetSettings(DomainName_String)
+            {
+                new FaucetSetting() { Name = "Email", Display = "E-mail", Type = EditorType.TextBox, Required = true },
+                new FaucetSetting() { Name = "Password", Display = "Password", Type = EditorType.Password, Required = true },
+            };
+        }
+    }
+
+
+    string DomainName_String = "https://FreeFaucet.io/";
```
cgerabit: probably we can introduce this as const since the value never changes

References:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/const

FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+
+
+    public override void Start()
+    {
+        //Title that shows in the browser. Is used to identify and close popup windows
+        Title = "Free Faucet";
+
+        //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
+        SuccessXPath = "//*";
+        FailXPath = "//*";
```
cgerabit: possible script improvement including success and fail xpath.

FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+	 if (!IsVisible(ExitButton))
+	 {
+		
+        //else go to normal login
+		GoToUrl(DomainName_String+"/login");
+        Wait();	
+	    Wait(2);		
+	
+        var box_user = ElementByXPath("//div[2]/div[2]/form/div[1]/input[@name='user_name']");
+        SetText(box_user, GetSetting("Email"));
```
cgerabit: Possible verification of whether the element exists could improve the user experience and facilitate debugging in the future.

FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+
+		 
+	 //if i NOT Login - go to Login Page
+	 if (!IsVisible(ExitButton))
+	 {
+		
+        //else go to normal login
+		GoToUrl(DomainName_String+"/login");
+        Wait();	
+	    Wait(2);
```
cgerabit: This does the same as Wait(3,4);

FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+
+        GoToUrl(DomainName_String+"/faucets");
+        Wait();		
+        Wait(2);
+		
+        var BitcoinCardHere = ElementByXPath("//div[6]/div/div[1]/h2[contains(text(),'Bitcoin') ]");
+        if (IsVisible(BitcoinCardHere))
+        {
+        Log("BitcoinCardHere - here");
+        var ClaimBitcoinStart = ElementByXPath("//div[6]/div/div[2]/div/div[1]/div[3]/form/input[3]");
```
cgerabit: if some xpath could be shortened, such as this one, it would be very beneficial for the utility life of the script.

FreeFaucet.io/FreeFaucet.io.fc
```diff
@@ -0,0 +1,269 @@
...
+//DoSolveFaucet
+//=============//
+
+
+
+
+    public override int AfterSolveFaucet()
+    {
+        return base.AfterSolveFaucet();
+    }
```
cgerabit: the method only returns the inherited by its base, it would be possible to eliminate it to improve the readability of the code.

References:

https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/inheritance

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/base

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/virtual

FreeFaucet.io/FreeFaucet.io.fc
faucetcollector: @aviramdayan can you check if you can find a proper fail/success

FreeFaucet.io/FreeFaucet.io.fc
faucetcollector: @aviramdayan please do add 
`if (!IsVisible(box_user))
{
   return Fail("User element not found");
}`

for all elements you search for.
Please also find a better xpath, not to use indexing like //div[2]/div[3] etc

FreeFaucet.io/FreeFaucet.io.fc
faucetcollector: Please find a proper xpath for this one, and all xpaths in the script.

FreeFaucet.io/FreeFaucet.io.fc
aviramdayan: no other possibility - in this specific html

FreeFaucet.io/FreeFaucet.io.fc
aviramdayan: no other possibility - in this specific html

FreeFaucet.io/FreeFaucet.io.fc
aviramdayan: is set in -  DoSolveFaucet

PR #71, opened 2019-11-17
Update-PopSpins.Com.fc

aviramdayan:
Fix_Spin_Element

### DIFF
--- a/PopSpins.Com.fc
+++ b/PopSpins.Com.fc
@@ -1,0 +1,488 @@
+
+//==================//
+//script Written by: ••AviramDayan•• 21/07/2019 18:15AM - Version 1 
+//Last Update: 17/11/2019 01:00AM - Version 14
+//==================//
+
+using System;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+//css_inc recaptchaUtilities
+
+public class popspins : FaucetScript
+{
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
+    /// You can get the value the user entered with the methods: GetSetting("[Name of the FaucetSetting]"), GetBoolSetting and GetDateTimeSetting
+    /// You can also create a new setting value using SetSetting("[Name you want to use]", "value")
+    /// </summary>
+    public override FaucetSettings Settings
+    {
+        get
+        {
+			
+            return new FaucetSettings("https://popspins.com/")
+            {
+                new FaucetSetting(){Name="Email",Display="Email or UserName",Type=EditorType.TextBox,Required=false},
+                new FaucetSetting(){Name="Password",Display="Password",Type=EditorType.Password,Required=false},
+                new FaucetSetting() { Name = "FaucetClaim", Display = "FaucetClaim - ON/OFF", Type = EditorType.CheckBox, Default = true},				
+                new FaucetSetting(){Name="StoreCookies",Display="StoreCookies - ON/OFF (also In delete browser data - i recommended to choose - default (delete after every run) Otherwise the script will not work properly)",Type=EditorType.CheckBox,Default=false },
+            };
+        }
+    }
+
+	
+//============================================================//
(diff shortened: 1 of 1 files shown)


PopSpins.Com.fc
```diff
@@ -0,0 +1,488 @@
...
+	        //=======//
+            var ClaimButton_2_Checking = ElementByXPath("//div[@class='claimouter' and contains(@style,'display: block;')]");
+            if(!IsVisible(ClaimButton_2_Checking)) //if the ClaimButton **NOT Visible** - only then call to do stuff - and return to exict time
+            {
+				
+            var MinuteVAR = ElementByXPath("//span[@id='countdown_time']");				
+            if(IsVisible(MinuteVAR))
+            {
+	        Log("MinuteVAR");  	
+            int MinuteINT = Convert.ToInt32(MinuteVAR.Text.Trim().Split(':')[1])*60;	 // need [4] for example if we have Existing text "You have to wait 15 minutes"  it mean = You[0] have[1] to[2] wait[3] 15[4] minutes[5]
```
faucetcollector: duplicate code from above, best to move it to a method

PopSpins.Com.fc
```diff
@@ -0,0 +1,488 @@
...
+	
+            //click one time before - while loop
+			Title = "POP Spins";
+            var Spin_one_Button = ElementByXPath("//input[@id='playFancy']");
+		    Click(Spin_one_Button);	
+            Wait(3);
+
+		    
+			//if the start button apper
+            while (true)
```
faucetcollector: There will be an endless loop if the spinbutton isn't found.

PopSpins.Com.fc
```diff
@@ -0,0 +1,488 @@
...
+	 //if i NOT Login - go to Login Page
+	 if (!IsVisible(ExitButton))
+	 {
+
+		GoToUrl("https://popspins.com/login.php");
+		Wait();		
+		Wait(2);
+	
+        var box_user = ElementByXPath("//input[@name='username']");
+        SetText(box_user, GetSetting("Email"));
```
faucetcollector: best to check if the element is found.

PopSpins.Com.fc
```diff
@@ -0,0 +1,488 @@
...
+	        //=======//
+            var ClaimButton_2_Checking = ElementByXPath("//div[@class='claimouter' and contains(@style,'display: block;')]");
+            if(!IsVisible(ClaimButton_2_Checking)) //if the ClaimButton **NOT Visible** - only then call to do stuff - and return to exict time
+            {
+				
+            var MinuteVAR = ElementByXPath("//span[@id='countdown_time']");				
+            if(IsVisible(MinuteVAR))
+            {
+	        Log("MinuteVAR");  
+            int MinuteINT = Convert.ToInt32(MinuteVAR.Text.Trim().Split(':')[1])*60;	 // need [4] for example if we have Existing text "You have to wait 15 minutes"  it mean = You[0] have[1] to[2] wait[3] 15[4] minutes[5]
```
faucetcollector: best to use int.TryParse

PopSpins.Com.fc
aviramdayan: in email/username & password area, it always found, 
the element from the owner side, change seldom ..

PopSpins.Com.fc
aviramdayan: I prefer like this way

PopSpins.Com.fc
aviramdayan: yeah but about the specific html/system if I move it to one function it clash things, so i prefer like this, in this case

PopSpins.Com.fc
aviramdayan: there is protection - return 1*60; in the elements/actions if not found

faucetcollector: That is in a different while loop. 
If it can't find the SpinButton then it will be in this while loop forever. It can happen already if they change the id or value/class. Or if you are not on the right page.

PR #72, opened 2019-11-17
Create CyberDice.net.fc

aviramdayan:


CyberDice.net/CyberDice.net.fc
```diff
@@ -0,0 +1,484 @@
...
+
+using System;
+using System.Drawing;
+using FaucetCollector.Script;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+
+//css_inc recaptchaUtilities
```
faucetcollector: do you need all these includes?

CyberDice.net/CyberDice.net.fc
```diff
@@ -0,0 +1,484 @@
...
+        var StartOpen_2_Login = ElementByXPath("//span[contains(text(),'Login')]");
+	    if (IsVisible(StartOpen_2_Login))
+	    { 
+        Click(StartOpen_2_Login);
+		}
+*/		
+		Wait(2);
+		
+        var box_user = ElementByXPath("//input[@name='auth-username']");
+        SetText(box_user, GetSetting("Email"));
```
faucetcollector: best to check if its visible or not

CyberDice.net/CyberDice.net.fc
```diff
@@ -0,0 +1,484 @@
...
+        Wait(2);
+
+        var StartOpenLogin = ElementByXPath("//a[@class='cta' and contains(text(),'Log in')] | //span[contains(text(),'Log In')]");
+	    if (IsVisible(StartOpenLogin))
+	    { 
+        Click(StartOpenLogin);
+		}
+
+		Wait(2);
+/*
```
faucetcollector: quite some disabled code in this .fc file

CyberDice.net/CyberDice.net.fc
aviramdayan: yeah for future work

CyberDice.net/CyberDice.net.fc
```diff
@@ -0,0 +1,484 @@
...
+        Click(Btc_Choose);
+        Log("click on Btc_Choose ");
+*/
+        Wait();	
+        Wait(2);
+
+        //cahck balance
+        var balance = ElementByXPath("//span[@class='header__wallet-value'] | //span[@class='user-wallet-balance']");
+        string BalanceSTRING = balance.Text.Trim().Split('.')[1];
+        int BalanceINT = Convert.ToInt32(BalanceSTRING);
```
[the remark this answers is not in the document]
aviramdayan: I prefer like this way

CyberDice.net/CyberDice.net.fc
aviramdayan: in email/username & password area, it always found, 
the element from the owner side, change seldom ..

CyberDice.net/CyberDice.net.fc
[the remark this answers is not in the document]
faucetcollector: then it will throw an exception when there isn't an int in the `BalanceSTRING`. Please always use `int.TryParse`

CyberDice.net/CyberDice.net.fc
faucetcollector: Please don't add them next time if they aren't used. If we ever change one of these includes we need to go over a lot of scripts that don't even use them.

CyberDice.net/CyberDice.net.fc
faucetcollector: Its not always found. It might be the website didn't load or they changed something. Its best to use defensive programming strategies. Please do it the next time.

PR #87, opened 2019-11-19
Add-Update-Geetest

aviramdayan:
Add-Update-Geetest


aviramdayan:
@cgerabit - let me know if something wrong here - and mention me why

faucetcollector:
The files should go into an `assets` subfolder, not into an `other` folder.
And you can only use the .zip file with the images, they will be auto extracted when FC downloads them. No need to put the .png files in there separately.

aviramdayan:
> The files should go into an `assets` subfolder, not into an `other` folder.
> And you can only use the .zip file with the images, they will be auto extracted when FC downloads them. No need to put the .png files in there separately.

I sent a new PR

PR #88, opened 2019-11-20
Add-Geetest-Updated-to-TrustDice.Win

aviramdayan:
Add-Geetest-Updated-to-TrustDice.Win


faucetcollector:
The zip file needs to be put in the `assets` folder, now its on the root

aviramdayan:
> The zip file needs to be put in the `assets` folder, now its on the root

weird, not i did this/or send it like that, i will try to create new again...

PR #89, opened 2019-11-21
Update Yannik.biz.fc

aviramdayan:
new-update-elements-and-way


Yannik.biz/Yannik.biz.fc
```diff
@@ -48,6 +48,390 @@ public class YannikScript : FaucetScript
...
+//==================//
+
+using FaucetCollector.Script;
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+//css_inc RecaptchaUtilities
+
+
+public class YannikScript : FaucetScript
```
faucetcollector: The script is duplicated (twice public class YannikScript : FaucetScript)

aviramdayan: weird, not i did this/or send it like that, i will try to create new again...

PR #90, opened 2019-11-21
Update-2-Yannik.biz

aviramdayan:
new-update-elements-and-way

### DIFF
--- a/Yannik.biz/Yannik.biz.fc
+++ b/Yannik.biz/Yannik.biz.fc
@@ -1,8 +1,16 @@
-﻿using FaucetCollector.Script;
+
+//==================//
+//script Written by: ••AviramDayan•• 14/06/2018 09:00AM - Version 1
+//Last Update: 22/11/2019 09:00AM - Version 28
+//==================//
+
+using FaucetCollector.Script;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
 //css_inc RecaptchaUtilities
+
+
 public class YannikScript : FaucetScript
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


Yannik.biz/Yannik.biz.fc
```diff
@@ -192,117 +229,110 @@ public class YannikScript : FaucetScript
...
+	    if (IsVisible(FaucetButton))
+	    { 
+        Click(FaucetButton);
+        Log("FaucetButton");
+		}
+		Wait();
+		Wait(2);
+
+        var Faucet_2_CallBackButton = ElementByXPath("//button[@data-callback='onSubmit']");
+	    if (IsVisible(Faucet_2_CallBackButton))
```
faucetcollector: If some of these elemente are not visible its best to return Fail. Otherwise it thinks its a success, but in fact it is not.

Yannik.biz/Yannik.biz.fc
```diff
@@ -30,159 +38,188 @@ public class YannikScript : FaucetScript
...
+        ad = false;
+       
+        //Title that shows in the browser. Is used to identify and close popup windows
         Title = "Yannik.biz";
 
-        //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
-        SuccessXPath = "//h4[contains(text(),'Captcha is wrong.')]";
-        FailXPath = "//div[contains(@class,'alert-danger')and contains(text(), 'Captcha is wrong.')]";
-
+		SuccessXPath = "//*";  //*[@id='']
```
faucetcollector: Why didn’t you add these, now it doesn’t log the claim success

Yannik.biz/Yannik.biz.fc
aviramdayan: after this step there is  SolveCaptcha_Function(); 
so if the recaptcha is popup/visible and not solve - is return to fail

Yannik.biz/Yannik.biz.fc
aviramdayan: again i prefer to not work with this methods
else i perfer to do protection Success/fail in the bottom - in DoSolveFaucet area

faucetcollector: But I don't see the claim success in here

Yannik.biz/Yannik.biz.fc
faucetcollector: No it doesn't - SolveCaptcha_Function only checks if recaptcha is visible.

PR #92, opened 2019-11-22
Update LuckyFish.io.fc

aviramdayan:
Update-Elements-Improvement


LuckyFish.io/LuckyFish.io.fc
```diff
@@ -47,6 +47,1539 @@ public class luckyfishorg : FaucetScript
                 new FaucetSetting() { Name = "bch", Display = "bch - claim", Type = EditorType.CheckBox, Default = true},
```
aviramdayan: Why the hell, that doubles the script

PR #94, opened 2019-11-22
Update-3-LuckyFish.io.fc

aviramdayan:
Update-3-Elements-Improvement

### DIFF
--- a/LuckyFish.io/LuckyFish.io.fc
+++ b/LuckyFish.io/LuckyFish.io.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 26/05/2019 04:00PM - Version 1
-//Last Update: 05/11/2019 21:00AM - Version 14
+//Last Update: 22/11/2019 11:00AM - Version 15
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


LuckyFish.io/LuckyFish.io.fc
```diff
@@ -327,31 +275,40 @@ public class luckyfishorg : FaucetScript
...
+		Log("Withdraw_2_Tab");
+		}	
 		
         Wait(2);
 
         //do all stuff	
         MenuCoins_Faunction(); //call
         Wait(2);
-        var btc_w_VAR = ElementByXPath("//div[6]/div[1]/div[1]/ul/li[1]/span");  
+        var btc_w_VAR = ElementByXPath("//ul[@class='el-scrollbar__view el-select-dropdown__list']/li[1]");
```
faucetcollector: Lot of same code, best would be to move most of it to a new method and only pass the xpath string.

LuckyFish.io/LuckyFish.io.fc
```diff
@@ -1115,29 +593,29 @@ public class luckyfishorg : FaucetScript
...
-			
+		}// GetBoolSetting - doge
+        //=========//		
+	
 		//======//
-		//omg//
+		//ltc//
 		//======//
-        if (GetBoolSetting("omg"))
+        if (GetBoolSetting("ltc"))
```
faucetcollector: Lot of same code, best would be to move to a new method and pass the setting string and xpath.

LuckyFish.io/LuckyFish.io.fc
```diff
@@ -1659,21 +1168,64 @@ public class luckyfishorg : FaucetScript
...
-        public void WithdrawEndButton_Faunction()
+     //=============//
+     //WithdrawEndButton_Faunction
+     //=============// 
+      public int WithdrawEndButton_Faunction()
+      {
+
+        var balance = ElementByXPath("//div[@class='available_balance_content']");
+        string BalanceSTRING = balance.Text.Trim().Split('.')[1];
+        int BalanceINT = Convert.ToInt32(BalanceSTRING);
```
faucetcollector: If it isnt an int this throws an exception and the script fails, best to use int.TryParse.

PR #95, opened 2019-11-23
Add Withdraw amount to settings

JReming85:
Converted static withdraw amount to intbox

Added to settings ```WithdrawAmt``` as a Numeric Text Box

Changed withdraw if higher then 37.5 to settings box (default 70)


Free-Dogecoin.com/Free-Dogecoin.com.fc
```diff
@@ -18,6 +21,7 @@ public class FreeDocoinScript : FaucetScript
                 new FaucetSetting(){Name="User",Display="Email",Type=EditorType.TextBox,Required=true},
                 new FaucetSetting(){Name="Pass",Display="Password",Type=EditorType.Password,Required=true},
                 new FaucetSetting(){Name="AutoWithdraw",Display="AutoWithdraw",Type=EditorType.CheckBox,Default=true},
+				new FaucetSetting(){Name="WithdrawAmt",Display="Withdraw Amount (Must include fees)",Type=EditorType.Numeric,Required=false,Default=70},
```
faucetcollector: Can you change the display into “Withdraw all when balance above” or something like that? Now it sounds like it only withdraws the amount you entered, but it does withdraw all

JReming85: fixed

PR #97, opened 2019-11-24
Update BestChange.com.fc

aviramdayan:
update-elements-and-time-options

### DIFF
--- a/BestChange.com/BestChange.com.fc
+++ b/BestChange.com/BestChange.com.fc
@@ -1,8 +1,7 @@
-﻿
 
 //==================//
 //script Written by: ••AviramDayan•• 13/09/2019 22:09PM - Version 1
-//Last Update: 13/09/2019 22:09PM - Version 1
+//Last Update: 24/11/2019 22:09PM - Version 2
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


BestChange.com/BestChange.com.fc
```diff
@@ -462,36 +459,61 @@ public class firebitcoinfaucettech : FaucetScript
...
-				
-         Driver.Close();  // Close specific chrome window
-         Driver.Quit();  // also Close the console app, that was used to kick off the chrome window
-         Driver.Dispose();  // Close the chromedriver.exe
-		 //return GetWaitSetting();
-         return MinuteINT;
+         var TimeVar = ElementByXPath("//input[@id='bonus_button']");
+
+         int min,sec;
+         min = Convert.ToInt32(TimeVar.Text.Trim().Replace("Time left: ", "").Split('.')[0])*60;
```
faucetcollector: I will keep asking to please use int.TryParse next time to improve script resilience

aviramdayan: but this method `Int32`, in all my scripts work well all the time..

faucetcollector: Yes, until something is changed and the value passed to ToInt32 isnt an int anymore then an Exception will be thrown and the script will fail.
So to improve resciliency its best to do int.TryParse so it doesnt throw an Exception but you can handle the case and return a custom time if it fails

aviramdayan: I'll run more tests later, thanks for the info.

PR #98, opened 2019-11-24
Update-3-Yannik.biz

aviramdayan:
new-update-elements-and-way-and-more-protection

### DIFF
--- a/Yannik.biz/Yannik.biz.fc
+++ b/Yannik.biz/Yannik.biz.fc
@@ -1,8 +1,16 @@
-﻿using FaucetCollector.Script;
+
+//==================//
+//script Written by: ••AviramDayan•• 14/06/2018 09:00AM - Version 1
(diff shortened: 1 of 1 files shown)


faucetcollector:
Next time please update your existing Pull Request (so update the same branch as before)
Now I cant see my old comments and what you have changed since the previous commit.
I still see int.Parse and no ClaimSuccess in here, but will merge now.

aviramdayan:
> Next time please update your existing Pull Request (so update the same branch as before)
> Now I cant see my old comments and what you have changed since the previous commit.
> I still see int.Parse and no ClaimSuccess in here, but will merge now.

from my experience here (in desktop web version) and not from app because there is problem with permissions
but anyway, if i create new `Commit` on the same branch
sometimes for me, it doubles the script, (i really don't know why it happens), so I prefer to create a new "branch".

PR #107, opened 2019-11-26
Create X6crypto.club (MySocial.Fun).fc

aviramdayan:
Create


faucetcollector:
Can you put it in the folder MySocial.Fun and name the .fc file the same?
You can mark the X6crypto.club as inactive then in the /admin.

aviramdayan:
> Can you put it in the folder MySocial.Fun and name the .fc file the same?
> You can mark the X6crypto.club as inactive then in the /admin.

yeah about the old I did inactive for all, and about the new name - that's what the owner wanted that all will knowes also, to not lose traffic probably,
so approve the new/last one

PR #110, opened 2019-11-29
Update Hitlink

cgerabit:
*New version of the script with the new version of the site


HitLink/Hitlink.fc
```diff
@@ -45,8 +46,8 @@ public class HitLink : FaucetScript
         Title = "HitLink";
 
         //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
-        SuccessXPath = "//frame[1]";
-        FailXPath = "//div[class='btn btn-red']";
+        SuccessXPath ="//font[contains(text(),'You must add a site below before surfing')]";
```
faucetcollector: Why are the fail and success the same?

HitLink/Hitlink.fc
```diff
@@ -274,18 +186,19 @@ public class HitLink : FaucetScript
...
                 }
                 Wait();
-                Driver.SwitchTo().Frame(ElementByXPath("//frame[1]"));
+                //Driver.SwitchTo().Frame(ElementByXPath("//frame[1]"));
                     Wait();
-                   // Log("SWITCH");
-                    while (IsVisible(ElementByXPath("//div[@id='timer']")))
+                // Log("SWITCH");
+                var Timer=ElementByXPath("//div[@id='timer']");
+                    while (IsVisible(Timer) && !Timer.Text.Contains("GO") )
```
faucetcollector: Can you add protection to do max attempts?

HitLink/Hitlink.fc
```diff
@@ -352,247 +264,110 @@ public class HitLink : FaucetScript
...
-                }
+            icons[i+1]=Principal_.Clone ( new Rectangle ( cont , 0 , 75 , 75 ) , Principal_.PixelFormat );
+           
+            cont+=75;
 
-            }
-   //         Log("Set solution");
+        }
+        for ( int i = 0 ; i<5 ; i++ )
+        {
```
faucetcollector: Does it save to disk? Could it use memory instead?

PR #113, opened 2019-12-02
Update Neobux.fc

aviramdayan:
fix-wait-protection-update-new-element-exists-copied-cg-dot-update-also

### DIFF
--- a/NeoBux/Neobux.fc
+++ b/NeoBux/Neobux.fc
@@ -1,4 +1,9 @@
-﻿using System;
+//============================================//
+//script Written first by: cgerabit 10/07/2018 14:00PM version 1
+//script Written last updates by: AviramDayan 02/12/2019 21:00PM version 19
+//============================================//
+
+using System;
 using System.Drawing;
 using System.IO;
 using FaucetCollector.Script;
@@ -240,7 +245,7 @@ public class CannonBux_ : FaucetScript
         Wait();
 
         var Ads = ElementsByXPath("//div[contains(@class,'cell c_adf')]//div[contains(@onmouseover,'h')]");
-       
+
         Log(Ads.Count + " Standard ads detected");
         try
         {
@@ -258,17 +263,20 @@ public class CannonBux_ : FaucetScript
                     {
                         CheckForPopupsEnabled = false;
                         Click(Ads[n], false);
-                        Wait();
+                        Wait(2);
+                        var AdsX2 = ElementByXPath("//div[contains(@class,'inB on')]/span[contains(@id,'da')]/a");
+                        Click(AdsX2);
+                        Wait(20);
                         var reddot = ElementsByXPath("//div[contains(@class,'cell c_adf')]//img[contains(@src,'imagens') and contains(@src,'dot')]");
                         if (IsVisible(reddot[n]))
                         {
                             Click(reddot[n]);
                         }
                         Driver.SwitchTo().Window(Driver.WindowHandles[1]);
-                        Wait(5);
+                        Wait(2);
                         bool breaked = false;
                        
-                        DateTime t2 = DateTime.Now.AddSeconds(20);
+                        DateTime t2 = DateTime.Now.AddSeconds(40);
                         while (IsVisible(ElementByXPath("//div[@id='omt1']//div[@class='msgDv2']")))
                         {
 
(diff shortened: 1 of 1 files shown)


NeoBux/Neobux.fc
```diff
@@ -258,17 +263,20 @@ public class CannonBux_ : FaucetScript
                     {
                         CheckForPopupsEnabled = false;
                         Click(Ads[n], false);
-                        Wait();
+                        Wait(2);
+                        var AdsX2 = ElementByXPath("//div[contains(@class,'inB on')]/span[contains(@id,'da')]/a");
+                        Click(AdsX2);
```
cgerabit: What is this ? and why you didn't checked if the element is visible before click it. In my side this element is not visible. 
please check that the element is visible before click it

NeoBux/Neobux.fc
```diff
@@ -258,17 +263,20 @@ public class CannonBux_ : FaucetScript
...
+                        Wait(20);
```
cgerabit: what is the reason to wait 20 seconds before clicking the reddot if you have not even started to see the ad? . 
20 seconds is a long time, please add the code to detect how long you should wait

NeoBux/Neobux.fc
```diff
@@ -282,7 +290,7 @@ public class CannonBux_ : FaucetScript
                         }
                         Wait();
                         DateTime t1 = DateTime.Now.AddSeconds(40);
-                        var errorbox = ElementByXPath("//div[@class='errorMn1']");
+                        var errorbox = ElementByXPath("//a[@class='errorMn1']");
```
cgerabit: In my side the element still being div not anchor , probably you can change that xpath to //*[@class='errorMn1']

NeoBux/Neobux.fc
```diff
@@ -377,12 +386,12 @@ public class CannonBux_ : FaucetScript
                     }
                     
               
-                    Wait(1);
+                    Wait(15);
```
cgerabit: Why wait 15 seconds? , the script have correctly the code to get the time that need to wait

NeoBux/Neobux.fc
```diff
@@ -294,7 +302,7 @@ public class CannonBux_ : FaucetScript
                         }
 
                         Wait();
-                        while (IsVisible(ElementByXPath("//canvas[@id='cnvs']"))){
+                        while (!IsVisible(ElementByXPath("//canvas[@id='cnvs' and contains(@style,'none')]"))){
```
cgerabit: Why you have inverted the condition? , "//canvas[@id='cnvs']" is the xpath of the element of the timer of every ad. if you invert the condition will wait while the timer is not visible i mean it will not wait.

NeoBux/Neobux.fc
aviramdayan: in the old script & html (and your last one) was only one time to click on some ad/row that are available, Right now when you click opens a certain row, 
the owner put more image to click & find taht in each row 
( the unique element on the way is `inB on' )
and the full is: `("//div[contains(@class,'inB on')]/span[contains(@id,'da')]/a")`
also no need to tell if is visible, because when you click he show the little image only one time
I noticed it, and you didn't notice it, i wouldn't put anything wrong,
anyway i put video/gif example to you

NeoBux/Neobux.fc
aviramdayan: ok this i agree - will do

NeoBux/Neobux.fc
aviramdayan: it's after not before - it's protection to add more some seconds to wait - because it switch & close too quickly

so this unique 
`("//div[contains(@class,'inB on')]/span[contains(@id,'da')]/a")`
and need to replace instead "reddot" image src that you set
`("//div[contains(@class,'cell c_adf')]//img[contains(@src,'imagens') and contains(@src,'dot')]")`
image

NeoBux/Neobux.fc
aviramdayan: i was wondering if i should do 10 or 15
but 15 more protects the user,
so I preferred to add extra time like this,
than it would fail, which is good.

NeoBux/Neobux.fc
cgerabit: Multiple dot elements have always existed since the first version of the script. 

 Click(Ads[n], false);
                        Wait();
                        var reddot = ElementsByXPath("//div[contains(@class,'cell c_adf')]//img[contains(@src,'imagens') and contains(@src,'dot')]");
                        if (IsVisible(reddot[n]))
                        {
                            Click(reddot[n]);
                        }

this code was responsible for clicking the correct dot, if we click the ad[n] , the correct dot is the reddot[n] (within the available ads)

PR #125, opened 2019-12-15
Update DiscordTip (Tip.cc).fc

aviramdayan:
general-all-improvement-update-all-elements-update-laccs-faucet

### DIFF
--- a/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
+++ b/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 05/12/2018 06:00AM 
-//Last Update: 13/12/2019 23:00PM - Version 52
+//Last Update: 15/12/2019 21:30PM - Version 53
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
```diff
@@ -1770,7 +1694,132 @@ public class discordapp_ : FaucetScript
       }
       //=================//	
 	  
+	
+      //=================//
+      //SetText_1_Laccs_SendKeys_Way_Function
+      //=================//	
+      public int SetText_1_Laccs_SendKeys_Way_Function()
```
faucetcollector: good job getting these into a different method.
you could even go one step further and make the value it needs to type ( `!claim` or `!faucet boost` etc) included as a parameter of one method.

aviramdayan: tnx dear 
you can put more example of what you mean? 
(for now i merged, but i want to know in general what you mean)

aviramdayan: for example i will explain something that i don't like
it's ridiculous to write & build all this for only one text "yes"
`
var TextCommand = ElementByXPath("//div[contains(@class,'innerEnabled')]/div[contains(@class,'textAreaEnabled')]");
	 //SetText Keyboard
         Log("TextCommand");
         Actions Builder4Act = new Actions(Driver);
         Builder4Act.MoveToElement(TextCommand).Build().Perform(); 
         Wait(3);
         Builder4Act.SendKeys(TextCommand, "yes").Build().Perform();
`
there is no method to build the whole process to declare with String,
(Instead of constantly rebuilding everything any time ..) 
then just write the all plain text, maybe there is an option?
`SendKeys(TextCommand, StringText).Build().Perform();`

faucetcollector: I was talking about something like this:
```
public int SendCommand(string command)
{
    var textArea = ElementByXPath("//div[contains(@class,'innerEnabled')]/div[contains(@class,'textAreaEnabled')]");
    if (!IsVisible(textArea))
    {
        return Fail("TextArea element not found");
    }
    Log("Send command" + command);
    var actions = new Actions(Driver);
    actions.MoveToElement(textArea).Build().Perform();
    Wait(3);
    actions.SendKeys(textArea, command).Build().Perform();
    return 0;
}
```

PR #126, opened 2019-12-16
Update LuckyFish.io.fc

aviramdayan:
Fix-Faucet-Elements-Menu-and-general

### DIFF
--- a/LuckyFish.io/LuckyFish.io.fc
+++ b/LuckyFish.io/LuckyFish.io.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 26/05/2019 04:00PM - Version 1
-//Last Update: 26/11/2019 11:00AM - Version 16
+//Last Update: 16/12/2019 03:00AM - Version 19
 //==================//
 
 using System;
@@ -67,7 +67,7 @@ public class luckyfishorg : FaucetScript
 				new FaucetSetting(){Name="MenuDefaultCoin",Display="MenuDefaultCoin - Select Favorite Coin to play OR for general stuff",Type=EditorType.ComboBox,Default="btc",Items=new System.Collections.Generic.List<string>{"btc","doge","ltc","eth","bch","xlm","xrp","dash","zec","usdt","bsv","btg","eos","link","bat","omg","hot","ht","usdc","tusd","pax"}},
 
                 new FaucetSetting(){Name="Slot_Amount",Display="Slot_Amount",Default = "0.00000001",Type=EditorType.TextBox,Required=false},
-                new FaucetSetting(){Name="Slot_Lines",Display="Slot_Lines",Default = "1",Type=EditorType.TextBox,Required=false},
+                new FaucetSetting(){Name="Slot_Lines",Display="Slot_Lines",Default = "2",Type=EditorType.TextBox,Required=false},
 
             };
         }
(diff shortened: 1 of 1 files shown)


LuckyFish.io/LuckyFish.io.fc
```diff
@@ -1503,25 +1503,27 @@ public class luckyfishorg : FaucetScript
...
-        BuilderAct.SendKeys(Keys.Space).Build().Perform();  // space
-        Log("ElementToEnter");
-		
-		Wait();		
-		Wait(2);	
+
+		Wait(5);	
+
+		Actions BuilderAct = new Actions(Driver);
+        BuilderAct.SendKeys(Keys.Space); //click space to spin
```
cgerabit: This will not do anything  you can try with:

Actions BuilderAct = new Actions(Driver);
BuilderAct.SendKeys ( Keys.Space ).Build ( ).Perform ( );

aviramdayan: tnx is work (i'll send new PR after)

LuckyFish.io/LuckyFish.io.fc
```diff
@@ -1503,25 +1503,27 @@ public class luckyfishorg : FaucetScript
...
         Log("Slot_Amount_VAR");		
 		
-		Wait(2);
+		Wait(1);
+
+        var AlretMessage_Minimum = ElementByXPath("//p[contains(text(),'Sorry,The min bet is')]");
+	    if (IsVisible(AlretMessage_Minimum)) //if already active 
+	    {
+		Actions Builder2Act = new Actions(Driver);
+        Builder2Act.SendKeys("d"); //click d in the keyboard to to set min bet
```
aviramdayan: @cgerabit How about that, it doesn't work also..
I tried also
` Builder2Act.SendKeys("d").Build().Perform();`  //click d in the keyboard to to set min bet

cgerabit: probably you need focus the element before send keys,

 Actions Builder2Act = new Actions(Driver);
        Builder2Act.MoveToElement ( ElementToHover ).Build ( ).Perform ( );
        Wait(1)
        Builder2Act.Click ( ).Build().Perform();
        Wait(1)
        Builder2Act.SendKeys ( "d" ).Build().Perform();

can you try that code (ElementToHover need to be the element in that you are trying to write)?

cgerabit: you probably even want to create a method,

 private void WriteInTheChat (string Text )
    {
        //this is only a example xpath you can replace the xpath for the textarea xpath in that you want to write
        var TextArea=ElementByXPath("//div[contains(@class,'textAreaEnabled')]"); 
        Actions Builder2Act = new Actions(Driver);
        Builder2Act.MoveToElement ( TextArea ).Build ( ).Perform ( );
        Wait ( 1 );
        Builder2Act.Click ( ).Build ( ).Perform ( );
        Wait ( 1 );
        Builder2Act.SendKeys ( Text ).Build ( ).Perform ( );
        Wait ( 1 );
    }

them you can use it like this in every site of script that you need to write in the chat:

    WriteInTheChat ( "d" );

So if in a future you need to change something of the code to write in the chat you will not need to change it a lot of times

other option is to use HumanActions ,  it should write in texarea without problems the way to use it is:
 private void WriteInTheChat (string Text )
    {
        HumanActions.DriverHumanActions actions=new HumanActions.DriverHumanActions(Driver);
        //this is only a example xpath you can replace the xpath for the textarea xpath in that you want to write
        var TextArea=ElementByXPath("//div[contains(@class,'textAreaEnabled')]");

        actions.SetText ( TextArea,Text );
    }
(if you will use this way remember to add //css_ref HumanActions.dll in the header of the script)

aviramdayan: yeah bro i saw before that there is a possibility if there is "multi-building"
to send to `string` (instead of rebuilding every time)
but here is not the case I do not need to support with "element"
because we sit on "canvas" and there is no element to support ..

So what's the way it will work on a regular keyboard (like Space, Enter, Alt, etc)

PR #139, opened 2019-12-22
Create CoinPayments.Net.fc

aviramdayan:
Create2

### DIFF
--- a/CoinPayments.Net/CoinPayments.Net.fc
+++ b/CoinPayments.Net/CoinPayments.Net.fc
@@ -1,0 +1,400 @@
+
+//==================//
+//script Written by: ••AviramDayan•• 22/12/2019 06:05AM - Version 1
+//Last Update: 22/12/2019 06:05AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+
+
+public class btcsamsity : FaucetScript
+{
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
+    /// You can get the value the user entered with the methods: GetSetting("[Name of the FaucetSetting]"), GetBoolSetting and GetDateTimeSetting
+    /// You can also create a new setting value using SetSetting("[Name you want to use]", "value")
+    /// </summary>
+    public override FaucetSettings Settings
+    {
+        get
+        {
+            return new FaucetSettings("https://CoinPayments.Net")
+            {
+                new FaucetSetting(){Name="Email",Display="UserName OR Email",Type=EditorType.TextBox,Required=true},
+                new FaucetSetting(){Name="Pass",Display="Password",Type=EditorType.Password,Required=true},
+				
+                new FaucetSetting(){Name="FaucetClaim",Display="FaucetClaim",Type=EditorType.CheckBox,Default=true},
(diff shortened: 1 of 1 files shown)


CoinPayments.Net/CoinPayments.Net.fc
```diff
@@ -0,0 +1,400 @@
...
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+
+
+public class btcsamsity : FaucetScript
```
faucetcollector: class name doesn't match the faucet

CoinPayments.Net/CoinPayments.Net.fc
```diff
@@ -0,0 +1,400 @@
...
+    {
+        get
+        {
+            return new FaucetSettings("https://CoinPayments.Net")
+            {
+                new FaucetSetting(){Name="Email",Display="UserName OR Email",Type=EditorType.TextBox,Required=true},
+                new FaucetSetting(){Name="Pass",Display="Password",Type=EditorType.Password,Required=true},
+				
+                new FaucetSetting(){Name="FaucetClaim",Display="FaucetClaim",Type=EditorType.CheckBox,Default=true},
+                new FaucetSetting(){Name="CancelSecurity",Display="Cancel Security - Like 2FA/Email",Type=EditorType.CheckBox,Default=true},
```
faucetcollector: shouldn't the customer do this them self?
you can't login if they have it enabled, so I don't see the use of this option

CoinPayments.Net/CoinPayments.Net.fc
```diff
@@ -0,0 +1,400 @@
...
+		
+        Wait(2);
+		
+        var box_pass = ElementByXPath("//input[@name='pass']");
+        SetText(box_pass, GetPassword("Pass"));
+		
+        Wait(2);
+	
+        //solve Captcha X2 XPath (1 = captcha image | 2 = input text )
+        var result = DoSolveImageCaptcha(ElementByXPath("//form[@class='form login-form']/img"),ElementByXPath("//input[@id='captcha_code']"));
```
faucetcollector: If the box_user or box_pass are not visible it will do attempt to do the captcha, while that will not work.
please check if IsVisible(box_pass) and IsVisible(box_user) above.

CoinPayments.Net/CoinPayments.Net.fc
```diff
@@ -0,0 +1,400 @@
...
+
+	  } //end CancelSecurity
+	
+	
+	
+	
+      //=============//
+      //FaucetClaim
+      //=============//
+      if (GetBoolSetting("FaucetClaim"))
```
faucetcollector: Are there times you don't want this? I think this option can be removed

CoinPayments.Net/CoinPayments.Net.fc
aviramdayan: yeah but there is more option like Disable Password - i'll keep

CoinPayments.Net/CoinPayments.Net.fc
aviramdayan: i will do new pr

CoinPayments.Net/CoinPayments.Net.fc
aviramdayan: i'll keep..

CoinPayments.Net/CoinPayments.Net.fc
aviramdayan: ok

PR #140, opened 2019-12-22
Create PlayRoyal.com.fc

aviramdayan:
Create

### DIFF
--- a/PlayRoyal.com/PlayRoyal.com.fc
+++ b/PlayRoyal.com/PlayRoyal.com.fc
@@ -1,0 +1,318 @@
+
+//==================//
+//script Written by: ••AviramDayan•• 22/12/2019 07:00AM - Version 1
+//Last Update: 22/12/2019 07:00AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+
+
+public class PlayRoyal : FaucetScript
+{
+    /// <summary>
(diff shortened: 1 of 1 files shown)


PlayRoyal.com/PlayRoyal.com.fc
```diff
@@ -0,0 +1,318 @@
...
+
+      //=============//
+      //FaucetClaim
+      //=============//
+      if (GetBoolSetting("FaucetClaim"))
+      {
+
+        //go to CancelSecurity FaucetClaim page
+		GoToUrl("https://playroyal.com/");
+        Log("wait 30 sec - to get daily token ");
```
faucetcollector: It just have to wait to get the claim?

PlayRoyal.com/PlayRoyal.com.fc
```diff
@@ -0,0 +1,318 @@
...
+//=============//
+//DoSolveFaucet
+//=============//
+  public override int DoSolveFaucet()
+  {
+
+      //=============//
+      //FaucetClaim
+      //=============//
+      if (GetBoolSetting("FaucetClaim"))
```
faucetcollector: I don't think anyone wants this option unchecked, you can remove it.

PlayRoyal.com/PlayRoyal.com.fc
aviramdayan: yeah

PlayRoyal.com/PlayRoyal.com.fc
aviramdayan: i keep

PR #143, opened 2019-12-22
Create CoinPayments.Net.fc

aviramdayan:
Create3


aviramdayan:
> Approved. Next time its best to update your existing PR - so I can see what your new changes where

i have problem with that I told you in the past..

PR #145, opened 2019-12-22
Create GigaBet.com.fc

aviramdayan:
Create2

### DIFF
--- a/GigaBet.com/GigaBet.com.fc
+++ b/GigaBet.com/GigaBet.com.fc
@@ -1,0 +1,339 @@
+
+//==================//
+//script Written by: ••AviramDayan•• 23/12/2019 01:35AM - Version 1
+//Last Update: 23/12/2019 01:35AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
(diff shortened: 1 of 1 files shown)


GigaBet.com/GigaBet.com.fc
```diff
@@ -0,0 +1,339 @@
...
+	
+        var StartLoginButton = ElementByXPath("//div[contains(@class,'hidden')]/button[contains(@data-modal-trigger,'log') and contains(@style,'font-size:14px')]");
+	    if (IsVisible(StartLoginButton))
+	    {
+		 Click(StartLoginButton);
+	    }
+        Wait();
+	    Wait(2);
+		
+        if (IsVisible(ElementByXPath ("//input[@id='modal--login--email']") ))
```
faucetcollector: I prefer to do it like this:
`var emailElement = ElementByXPath ("//input[@id='modal--login--email']");
 if (!IsVisible(emailElement)) 
 {
     return Fail("Email element not found");
 }   
SetText(emailElement, GetSetting("Email"));
`

aviramdayan: from my experience this always do some clash, i prefer without

PR #148, opened 2019-12-25
Update GetZen.Cash.fc

aviramdayan:
changed-from-RecaptchaUtilities-to-normal-and-update-SuccessFailXPath

### DIFF
--- a/GetZen.Cash/GetZen.Cash.fc
+++ b/GetZen.Cash/GetZen.Cash.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 23/12/2019 01:30PM - Version 1
-//Last Update: 23/12/2019 01:30PM - Version 1
+//Last Update: 25/12/2019 03:40PM - Version 2
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


GetZen.Cash/GetZen.Cash.fc
```diff
@@ -295,24 +295,37 @@ public class GetZen : FaucetScript
...
+
+        var SuccessXPath = ElementByXPath("//div[@class='modal-content']//div[@id='result1' and contains(text(),'This address claimed Zen in the last')]");       
+		if (!IsVisible(SuccessXPath))
+        {
+         Log("success - claim");		
+         //ad = true; //success
+		 ClaimSuccess(); //Marks the claim as a success
+		}
+
+        if (IsVisible(FailXPath))
```
faucetcollector: This won't work, as FailXPath is a string, not a IWebElement.
Above with SuccessXPath it works, as you create a local one with an element.

aviramdayan: Oh ... i knew that maybe i have forgotten to fix it 
Instead `if (IsVisible(FailXPath))`
it should have been that way `if (IsVisible(SuccessXPath))` <
i send new pr again

PR #170, opened 2020-01-11
Update BestChange.com.fc

aviramdayan:
fix-Recaptcha-add-iframe

### DIFF
--- a/BestChange.com/BestChange.com.fc
+++ b/BestChange.com/BestChange.com.fc
@@ -2,7 +2,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 13/09/2019 22:09PM - Version 1
-//Last Update: 25/12/2019 05:10PM - Version 6
+//Last Update: 11/01/2020 15:40PM - Version 7
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


BestChange.com/BestChange.com.fc
```diff
@@ -352,37 +304,33 @@ public class firebitcoinfaucettech : FaucetScript
...
+		   var ReCaptchaButtonStart = ElementByXPath("//div[@class='rc-anchor-logo-portrait']");  //different
 	       if (IsVisible(ReCaptchaButtonStart))
 	       { 
            Click(ReCaptchaButtonStart);
 		   Log("ReCaptchaButtonStart - faucet");
+		   	   
+             var SuccessMessageExist = ElementByXPath("//div[@id='bonus_captcha_title' and contains(text(),'successfully')]");
+	         if (IsVisible(SuccessMessageExist))
+	         {
+	            Driver.SwitchTo().DefaultContent(); //go back to default content in <html>
```
faucetcollector: Shouldn't it do this line always when its finished in the iframe, not only in these two if statements?

aviramdayan: yeah only like this way ..
and in this section only > SwitchToIframe (no in login area)

PR #177, opened 2020-01-17
Allow referral code on first run

JReming85:
Add a bool to use a user defined referral code that will self disable after use


MoonBitco.in/MoonBitco.in.fc
```diff
@@ -96,6 +99,11 @@ public class MoonBitcoinScript : FaucetScript
     /// </summary>
     public override int DoLogin()
     {
+    	if (GetBoolSetting("useref"))
+		{
+			GoToUrl("http://moonbit.co.in/?ref=" + GetSetting("refcode"));
+			SetSetting("useref", false);
+	}
```
cgerabit: this will not work, the script will go to the site in DoInit method then as the script has already entered to the site 
so go to the url with the referral will not make any changes

JReming85: it works fine on moonbitcoin but I understand your concern and agree it should not work on the ones that require a button press before email entry.

I will move it to DoInit

PR #190, opened 2020-01-23
Update claimbits.net - shortlinks.fc

cgerabit:
*change xpath in login section

### DIFF
--- a/Claimbits Shortlinks/claimbits.net - shortlinks.fc
+++ b/Claimbits Shortlinks/claimbits.net - shortlinks.fc
@@ -91,13 +91,13 @@ public class ClaimbitsShortlink: FaucetScript
     /// <returns></returns>
     public override int DoLogin()
     {
-        var UserLogin = ElementByXPath("//input[@id='userLogin']");
+        var UserLogin = ElementByXPath("//input[@placeholder='Username / Email']");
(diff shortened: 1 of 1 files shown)


Claimbits Shortlinks/claimbits.net - shortlinks.fc
```diff
@@ -91,13 +91,13 @@ public class ClaimbitsShortlink: FaucetScript
     /// <returns></returns>
     public override int DoLogin()
     {
-        var UserLogin = ElementByXPath("//input[@id='userLogin']");
+        var UserLogin = ElementByXPath("//input[@placeholder='Username / Email']");
```
aviramdayan: Set to placeholder, this is sometimes a risk ..
Because it can change quickly from the owner, 
set to set to others Attributes/elements Better
But it is possible to do like this way, it will be better
And shorten the uppercase problems
//email
`
("//form[contains(@onsubmit,'loginAjax')]//input[contains(@placeholder,'name')] | //form[contains(@onsubmit,'loginAjax')]//input[contains(@placeholder,'mail')]")
`
//password
`
("//form[contains(@onsubmit,'loginAjax')]//input[contains(@placeholder,'word')]")
`

PR #207, opened 2020-02-01
Create Freegridco.in.fc

aviramdayan:
create

### DIFF
--- a/Freegridco.in/Freegridco.in.fc
+++ b/Freegridco.in/Freegridco.in.fc
@@ -1,0 +1,401 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 01/02/2020 10:30AM - Version 1
+//Last Update: 01/02/2020 10:30AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+
+
+public class btcsamsity : FaucetScript
+{
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
+    /// You can get the value the user entered with the methods: GetSetting("[Name of the FaucetSetting]"), GetBoolSetting and GetDateTimeSetting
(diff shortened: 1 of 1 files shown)


Freegridco.in/Freegridco.in.fc
```diff
@@ -0,0 +1,401 @@
...
+
+        if (IsVisible(ElementByXPath ("//input[@name='login']") )) 
+	    {
+         var box_user = ElementByXPath("//input[@name='login']");
+         SetText(box_user, GetSetting("Email"));
+        }
+		
+		Wait(2);
+		
+        if (IsVisible(ElementByXPath ("//input[@name='password']") ))
```
faucetcollector: Best to do the elementbyxpath only once. Also if you dont log or exit then you dont need the if statement at all, as the settext will not do anything if the element is not found

Freegridco.in/Freegridco.in.fc
```diff
@@ -0,0 +1,401 @@
...
+	{
+        ad=false; //solve the problem in start - after he return to script again
+        //return GetWaitSetting();  
+
+        var WaitTimerText = ElementByXPath("//p[@id='roll_wait_text']");	
+        if (IsVisible(WaitTimerText))
+        {
+             Log("FaucetClaim: WaitTimerText apper"); 
+             int min,sec;
+             min = Convert.ToInt32(WaitTimerText.Text.Trim().Replace("Wait for", "").Replace("before next roll", "").Split(':')[0])*60;
```
faucetcollector: Please use int.TryParse

Freegridco.in/Freegridco.in.fc
```diff
@@ -0,0 +1,401 @@
...
+             sec = Convert.ToInt32(WaitTimerText.Text.Trim().Replace("Wait for", "").Replace("before next roll", "").Split(':')[1]);
```
faucetcollector: Best to put the result of the .Split into a variable and use that to get the sec and min calculated

aviramdayan: yeah also way, but it work, in the next time, tnx for the tip

PR #212, opened 2020-02-06
Update MobileBet.com.fc

aviramdayan:
fix-elements

### DIFF
--- a/MobileBet.com/MobileBet.com.fc
+++ b/MobileBet.com/MobileBet.com.fc
@@ -1,8 +1,8 @@
-﻿
+
 
 //==================//
 //script Written by: ••AviramDayan•• 23/25/2019 21:09PM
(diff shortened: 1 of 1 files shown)


MobileBet.com/MobileBet.com.fc
```diff
@@ -114,7 +114,50 @@ public class mobilebet : FaucetScript
...
+		
+        Recaptcha_Function(); //call
+		
+		Wait(4);
+			
+	    var AccountClosed = ElementByXPath("//p[contains(text(),'closed') or contains(text(),'Your account is') ]");
+	    if (IsVisible(AccountClosed))
+	    { 
+		Log("AccountClosed - we return after 450 min");
+        Driver.Close();  // Close specific chrome window
```
faucetcollector: I see you had it before in the old script - but better let FC handle the driver management.

PR #224, opened 2020-02-17
Create DuckDice.io.fc

aviramdayan:
create

### DIFF
--- a/DuckDice.io/DuckDice.io.fc
+++ b/DuckDice.io/DuckDice.io.fc
@@ -362,12 +362,12 @@ public class btcsamsity : FaucetScript
 		  {
             //SolveCaptcha if apper
             var result = base.DoSolveCaptcha();
-            RecaptchaUtility utility = new RecaptchaUtility(this);
-            utility.DoSolve();
             if (result > 0)
             {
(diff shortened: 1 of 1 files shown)


DuckDice.io/DuckDice.io.fc
```diff
@@ -0,0 +1,378 @@
...
+          if (IsVisible(RecaptchaOpen_Iframe))
+		  {
+            //SolveCaptcha if apper
+            var result = base.DoSolveCaptcha();
+            RecaptchaUtility utility = new RecaptchaUtility(this);
+            utility.DoSolve();
+            if (result > 0)
+            {
+                return Fail("Fail solving captcha");
+            }
```
cgerabit: Probably  you will need to organize it this way

`var result = base.DoSolveCaptcha();
            
            if (result > 0)
            {
                return Fail("Fail solving captcha");
            }
RecaptchaUtility utility = new RecaptchaUtility(this);
  utility.DoSolve();`

Like that if recaptcha fail you will not call the callback

aviramdayan: ok before..
thanks for tip (we take to next created scripts also)

PR #248, opened 2020-03-04
Update my.dropz.xyz.fc

aviramdayan:
add-AutoSurf-AutoWithdraw

### DIFF
--- a/my.dropz.xyz/my.dropz.xyz.fc
+++ b/my.dropz.xyz/my.dropz.xyz.fc
@@ -35,7 +35,7 @@ public class mydropzxyz : FaucetScript
                 new FaucetSetting(){Name="FaucetClaim",Display="FaucetClaim",Type=EditorType.CheckBox,Default=true},		
                 new FaucetSetting(){Name="AutoSurfAds",Display="AutoSurfAds",Type=EditorType.CheckBox,Default=true},
 
-                new FaucetSetting(){Name="AutoWithdraw",Display="AutoWithdraw",Type=EditorType.CheckBox,Default=true},
+                new FaucetSetting(){Name="AutoWithdraw",Display="AutoWithdraw",Type=EditorType.CheckBox,Default=false},
                 new FaucetSetting() { Name = "LimitWithdraw", Display = "LimitWithdraw", Type = EditorType.Numeric, Default = 3000 },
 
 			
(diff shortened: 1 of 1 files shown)


my.dropz.xyz/my.dropz.xyz.fc
```diff
@@ -264,19 +268,129 @@ public class mydropzxyz : FaucetScript
...
+			 
+           GoToUrl("https://my.dropz.xyz/account/");
+		   Wait(2);
+		   
+           var balance = ElementByXPath("//b[@id='display_pending_drops']");
+           string BalanceSTRING = balance.Text.Trim().Replace(",", "").Split('.')[0];
+           int BalanceINT = Convert.ToInt32(BalanceSTRING);
+   
+           int MyAmount;
+           Int32.TryParse(GetSetting("LimitWithdraw"), out MyAmount);
```
cgerabit: In these cases tryparse is always better, if the sintax of the balance change the script will stop work 
The better way to do it is:
if( Int32.TryParse(GetSetting("LimitWithdraw"), out MyAmount) &&  Int32.TryParse(BalanceSTRING, out BalanceINT )){
// Withdrawal code

}

so if everything goes wrong the other parts of the script will continue to work

my.dropz.xyz/my.dropz.xyz.fc
```diff
@@ -264,19 +268,129 @@ public class mydropzxyz : FaucetScript
...
+           Wait();	
+	       Wait(2);	
+			
+           var AutoSurf_Check = ElementByXPath("//i[@class='fa fa-play'] | //div[@id='adcopy-puzzle-image'] | //iframe[contains(@src,'recaptcha/api2')]");
+	       if (IsVisible(AutoSurf_Check))
+	       {
+			   
+		     Log("Bonus_Aff - Available");
+             var TimeToWaitForBreak = DateTime.Now.AddSeconds(65);
+             while (true)
```
cgerabit: while true is a bad practice maybe you can put while(IsVisible(AutoSurft_Check)) or even put the break in the while while(DateTime.Now > TimeToWaitForBreak && IsVisible(AutoSuft_Check))  (remember update autosuft check using: AutoSurf_Check = ElementByXPath("//i[@class='fa fa-play'] | //div[@id='adcopy-puzzle-image'] | //iframe[contains(@src,'recaptcha/api2')]"); if you want to do this)

my.dropz.xyz/my.dropz.xyz.fc
aviramdayan: i prefer to work with this method, for now.

my.dropz.xyz/my.dropz.xyz.fc
aviramdayan: it works just fine, there are many protections, 
i just will send new commit to set the auto Withdraw to false.

my.dropz.xyz/my.dropz.xyz.fc
```diff
@@ -264,19 +268,97 @@ public class mydropzxyz : FaucetScript
...
+               var FaucetButton = ElementByXPath("//input[@type='submit' and @value='Submit']");
+               Click(FaucetButton);
+			   Wait(2);
+			   Driver.Navigate().Refresh(); //refresh page
+			   }
+			   
+               Driver.SwitchTo().Window(Driver.WindowHandles[0]);	
+			   
+               var ButtonToWait = ElementByXPath("//a[@class='btn btn-app' and contains(text(),'Please wait')]");
+               while (IsVisible(ButtonToWait))
```
faucetcollector: This can result in an endless loop.
Also the indentation is a bit off - hard to see where thise while loop ends

my.dropz.xyz/my.dropz.xyz.fc
```diff
@@ -307,12 +389,54 @@ public class mydropzxyz : FaucetScript
 		  
           if (!IsVisible(FaucetCheck))
 	      {
+		  Log("FaucetClaim - break - not Available");
 		  break;
 		  }
-		  
+
+          if (DateTime.Now > TimeToWaitForBreak) //if the time above 65 sec 
+          {
```
faucetcollector: 👍

my.dropz.xyz/my.dropz.xyz.fc
aviramdayan: but there is `DateTime.Now.AddSeconds` in general that handle for all section..

aviramdayan: i will send more commit with `DateTime.Now.AddSeconds` inside this section ^

PR #249, opened 2020-03-04
Update BitsFreeScript.fc

cgerabit:
*try close popup in bitsfree

### DIFF
--- a/BitsFree/BitsFreeScript.fc
+++ b/BitsFree/BitsFreeScript.fc
@@ -242,16 +242,17 @@ public class BitsFreeScript_: FaucetScript
                 try
                 {
                                                         
-
-                                     
+                      
+                   
                       IconCaptchaSolver solver=new IconCaptchaSolver(this);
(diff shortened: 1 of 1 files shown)


BitsFree/BitsFreeScript.fc
```diff
@@ -242,10 +242,19 @@ public class BitsFreeScript_: FaucetScript
...
-                       Wait ();
+
+                                     
                       IconCaptchaSolver solver=new IconCaptchaSolver(this);
-                    if ( solver.IsVisibleIconCaptcha ( ) )
+                    Click ( Shortlinks[NShort] );
+                    Wait ( );
+                    while ( IsVisible ( Shortlinks[NShort] )&& !solver.IsVisibleIconCaptcha() )
+                    {
+                        Click ( Shortlinks[NShort] );
```
faucetcollector: If the click aint working due to whatever reason - and the element stays visible - this will result in an endless loop.

PR #250, opened 2020-03-04
Update doge-bitsfree.net.fc

cgerabit:
*Fix popup in the shortlink click

### DIFF
--- a/doge-bitsfree.net/doge-bitsfree.net.fc
+++ b/doge-bitsfree.net/doge-bitsfree.net.fc
@@ -247,10 +247,12 @@ public class Dogebitsfree: FaucetScript
                     IconCaptchaSolver solver=new IconCaptchaSolver(this);
                     Click ( Shortlinks[NShort] );
                     Wait ( );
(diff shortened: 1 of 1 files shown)


doge-bitsfree.net/doge-bitsfree.net.fc
```diff
@@ -235,20 +235,29 @@ public class Dogebitsfree: FaucetScript
...
-                      IconCaptchaSolver solver=new IconCaptchaSolver(this);
+
+
+
+                    IconCaptchaSolver solver=new IconCaptchaSolver(this);
+                    Click ( Shortlinks[NShort] );
+                    Wait ( );
+                    while ( IsVisible ( Shortlinks[NShort] )&&!solver.IsVisibleIconCaptcha ( ) )
+                    {
+                        Click ( Shortlinks[NShort] );
```
faucetcollector: If the click aint working due to whatever reason - and the element stays visible - this will result in an endless loop.

cgerabit: fixed

PR #257, opened 2020-03-08
version 1.0.0

avril-verhaeghen:
First version for the faucet CoinAdster

### DIFF
--- a/CoinAdster/CoinAdster.fc
+++ b/CoinAdster/CoinAdster.fc
@@ -30,7 +30,7 @@ public class CoinAdsterScript : FaucetScript
     {
         OverriteSolvemedia = false;
         //Title that shows in the browser. Is used to identify and close popup windows
-        Title = "CoinAdster - more than just faucet!";
+        Title = "CoinAdster";
 
         //After we did try to claim on the faucet we search for these elements to determine if it was a success or a fail
         SuccessXPath = "//div[@id='faucetMessage']//div[@class='alert alert-success']";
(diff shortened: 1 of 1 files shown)


CoinAdster/CoinAdster.fc
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /// <summary>
+    /// This method gets called when the faucet is enabled and the GO button is clicked
+    /// </summary>
+    public override void Start()
+    {
+        OverriteSolvemedia = false;
+        //Title that shows in the browser. Is used to identify and close popup windows
+        Title = "CoinAdster - more than just faucet!";
```
faucetcollector: You can also just use `"CoinAdster"`, as the popup check will do use a partitial match (`Contains`)

CoinAdster/CoinAdster.fc
```diff
@@ -0,0 +1,418 @@
...
+    {
+        //In here we can check for the presence of an element that is only present when you are logged in.
+        //In this case we check for en element with the class "loggedIn"
+
+        //If there is the user info block element then we are logged in - that element is only present when you are logged in.
+        var isLoggedIn = ElementByXPath("//div[@class='user']//span[@class='logout']") != null;
+        if (isLoggedIn)
+        {
+            //We are logged in, try to disable the sounds of the faucet.
+            ExecuteScript("faucetVM.alertPlaySound(false)");
```
faucetcollector: The moon faucets have this javascript available - is this one the same? Probably not - so better to remove this one.

CoinAdster/CoinAdster.fc
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /// <summary>
+    /// This method gets called after BeforeSolveCaptcha and it does the actual captcha solving
+    /// It will automatically try to see if there is a SolveMedia captcha active or a reCAPTCHA active.
+    /// If it finds either one it will then use the selected method on the Captcha tab of FaucetCollector to solve the Captcha.
+    /// </summary>
+    public override int DoSolveCaptcha()
+    {
+        if (GetBoolSetting("UseSolveMedia") && !OverriteSolvemedia)
```
faucetcollector: CoinAdster supports switching between solvemedia and recapthca - but you left out that option. That is fine for the first version, but better remove the code that you are not using.

PR #263, opened 2020-03-11
Update EarnBtc.io.fc

aviramdayan:
update-fix-and-more

### DIFF
--- a/EarnBtc.io/EarnBtc.io.fc
+++ b/EarnBtc.io/EarnBtc.io.fc
@@ -1,8 +1,8 @@
-﻿
+
 
 //==================//
 //script Written by: ••AviramDayan•• 26/06/2019 00:30PM - Version 1
-//Last Update: 01/09/2019 00:30PM - Version 2
+//Last Update: 11/03/2020 09:30PM - Version 7
 //==================//
 
 using System;
@@ -10,7 +10,6 @@ using FaucetCollector.Script;
 using OpenQA.Selenium;
 using OpenQA.Selenium.Interactions;
 using System.Text.RegularExpressions;
-using System.Linq;
 //css_inc RecaptchaUtilities
 //css_inc ShortLinkUtilities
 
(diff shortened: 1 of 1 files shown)


EarnBtc.io/EarnBtc.io.fc
```diff
@@ -219,25 +213,30 @@ public class EarnBtc : FaucetScript
...
-        }
+         if(IsVisible(Time_Min_Checking_Exist)) //Visible
+         {
+		 
+         Log("CountDownTimer");  
+
+         var MinVAR = ElementByXPath("//span[@id='minutes']"); 
+         var SecVAR = ElementByXPath("//span[@id='seconds']");
+		 
+         int MinINT = Convert.ToInt32(MinVAR.Text)*60;
```
faucetcollector: `int.TryParse`

EarnBtc.io/EarnBtc.io.fc
```diff
@@ -290,46 +282,195 @@ public class EarnBtc : FaucetScript
...
-        }
-       
-        Wait();
-        var GoToClaim = ElementByXPath("//button[@class='btn btn-primary btn-lg']");
-        if (!IsVisible(GoToClaim))
+         GoToUrl("https://earnbtc.io/dashboard");
+         Wait();
+         Wait(2);
+
+	     AutoWithdraw_Function(); //call
```
faucetcollector: Preferably we don't do auto withdraw in scripts anymore, to prevent multi botting

EarnBtc.io/EarnBtc.io.fc
```diff
@@ -290,46 +282,195 @@ public class EarnBtc : FaucetScript
...
+
+         if(IsVisible(Time_Min_Checking_Exist)) //Visible
+         {
+		 
+         Log("CountDownTimer");  
+
+         var MinVAR = ElementByXPath("//span[@id='minutes']"); 
+         var SecVAR = ElementByXPath("//span[@id='seconds']");
+		 
+         int MinINT = Convert.ToInt32(MinVAR.Text)*60;
```
faucetcollector: `int.TryParse` :)

aviramdayan: No i don't want to...
i like this way leave it like that, enough .. hehe

EarnBtc.io/EarnBtc.io.fc
aviramdayan: why? But I do limit

faucetcollector: because it makes it easier for people to use multiple accounts. they can just do 1 account and manual withdraw.

PR #264, opened 2020-03-11
Faucet waiting time bug fix

avril-verhaeghen:


avril-verhaeghen:
> Hello, Thanks for sending and updating the script,
> But unfortunately, from my experience i think you should stay away from this system
> (I've done a lot of scripts similar to this one)
> This system probably does not fit the bot, sooner or later, the balance is reset.
> So it's a waste of time to update, or make a new one.

Thanks for your reply, but can you elaborate? I'm not sure if I understand what you mean. Do you mean the script in it's whole or the last update that is checking for a countdown?

aviramdayan:
> > Hello, Thanks for sending and updating the script,
> > But unfortunately, from my experience i think you should stay away from this system
> > (I've done a lot of scripts similar to this one)
> > This system probably does not fit the bot, sooner or later, the balance is reset.
> > So it's a waste of time to update, or make a new one.
> 
> Thanks for your reply, but can you elaborate? I'm not sure if I understand what you mean. Do you mean the script in it's whole or the last update that is checking for a countdown?

PR #265, opened 2020-03-11
Update doge-bitsfree.net.fc

cgerabit:
*Add waiting for the button
*Update the xpath for the button and use modal instead of button text


aviramdayan:
unfortunately, from my experience i think you should stay away from this system
(I've done a lot of scripts similar to this one) and you also,
This system probably does not fit the bot, sooner or later, the balance is reset.
So it's a waste of time to update, or make a new one.

PR #266, opened 2020-03-12
Create FreeDogeonScript

AnonymousDeveloper65535:
This site has no captcha so you can easily automate it

---
This change is [Reviewable](https://reviewable.io/reviews/faucet-collector/scripts/266)

### DIFF
--- a/freedogeon.com/FreeDogeonScript
+++ b/freedogeon.com/FreeDogeonScript
@@ -30,7 +30,7 @@ public override int DoLogin()
 
 public override void Start()
 {
-    Title = "FreeDogeon.com - Win free dogecoins every hour!|Free Dogecoin Every Hour";
+    Title = "FreeDogeon.com|Free dogecoin";
 
     SuccessXPath = "//*[@id='free_play_result']";
     FailXPath = "//*[@id='free_play_error']";
--- a/freedogeon.com/FreeDogeonScript
+++ b/freedogeon.com/FreeDogeonScript
@@ -1,0 +1,96 @@
+using FaucetCollector.Script;
+using System;
+public class FreeDogeonScript : FaucetScript
+{
+public override int DoLogin()
+{
+    OpenQA.Selenium.IWebElement signInEmail = ElementById("login_form_doge_address");
+    if (signInEmail == null)
+    {
+        return Fail("Signin email input not found.");
+        
+    }
+    OpenQA.Selenium.IWebElement signInPassword = ElementById("login_form_password");
+    if (signInPassword == null)
+    {
+        return Fail("Signin password input not found.");
+        }
+    SetText(signInEmail, GetSetting("Email"));
+    SetText(signInPassword, GetPassword("Password"));
+    var button = ElementById("login_button");
+    if (!IsVisible(button))
+    {
+       return Fail("Login button not found");
+       
+    }
+
+    Click(button);
+    return base.DoLogin();
+}
+
+public override void Start()
+{
+    Title = "FreeDogeon.com - Win free dogecoins every hour!|Free Dogecoin Every Hour";
+
+    SuccessXPath = "//*[@id='free_play_result']";
+    FailXPath = "//*[@id='free_play_error']";
+    base.Start();
+}
+ 
+
+public override FaucetSettings Settings
+{
+    get
+    {
+        return new FaucetSettings("https://freedogeon.com/")
+        {
+            new FaucetSetting() {Name = "Email", Display = "Email", Type = EditorType.TextBox, Required = true},
+
+            new FaucetSetting() { Name = "Password", Display = "Password", Type = EditorType.Password, Required = true }
+        };
+    }
+}
+ public override int DoSolveFaucet()
+    {
+        //Find the submit button to claim on the faucet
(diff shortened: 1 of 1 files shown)


aviramdayan:
i saw this domain, the design was taken from "freedoge.co.in" 
So need a little searching and checking. 
i'm not reviewing the script right now, just writing in general.

AnonymousDeveloper65535:
ill be adding code to solve the captcha bc i just found out every once in a while a captcha will show up

freedogeon.com/FreeDogeonScript
```diff
@@ -0,0 +1,103 @@
...
+       
+    }
+
+    Click(button);
+    return base.DoLogin();
+}
+
+public override void Start()
+{
+    Title = "FreeDogeon.com - Win free dogecoins every hour!|Free Dogecoin Every Hour";
```
faucetcollector: FC uses a partitial match checking what popups to close. So you could do `Title = "FreeDogeon.com|Free dogecoin";`

freedogeon.com/FreeDogeonScript
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        string minutes = Convert.ToString(ExecuteScript("return $('#time_remaining .countdown_amount')[0].innerText"));
+        string seconds = Convert.ToString(ExecuteScript("return $('#time_remaining .countdown_amount')[1].innerText"));
+        if (string.IsNullOrEmpty(seconds) || string.IsNullOrEmpty(minutes))
+        {
+            return 0;
+        }
+
+        int min = Convert.ToInt32(minutes.Trim());
```
faucetcollector: Preferably use int.TryParse, for example:
```
int minutes;
if (!int.TryParse(minutes.Trim(), out minutes))
{
  //unable to parse.
  return base.GetFaucetWaitTime();
}
```

freedogeon.com/FreeDogeonScript
```diff
@@ -0,0 +1,103 @@
...
+            new FaucetSetting() {Name = "Email", Display = "Email", Type = EditorType.TextBox, Required = true},
+
+            new FaucetSetting() { Name = "Password", Display = "Password", Type = EditorType.Password, Required = true }
+        };
+    }
+}
+ public override int DoSolveFaucet()
+    {
+        //Find the submit button to claim on the faucet
+        OpenQA.Selenium.IWebElement submitButton = ElementByXPath("//*[@id='free_play_tab']/div[4]/div/p/button");
```
faucetcollector: Specifying the exact number of the element position in an array is not best practice, as when they change around the html its easy to break. This one aint that bad tho. So just a note, not really required to fix.

faucetcollector:
Thanks @AnonymousDeveloper65535 for taking the time on doing a PR with this script. As aviram pointed out - its an exact copy of freedoge.co.in - so its a bit suspicious if it turns out to be scam or not.

freedogeon.com/FreeDogeonScript
AnonymousDeveloper65535: like this?

faucetcollector: no, the int.TryParse will return a bool - if it could parse the input to an integer or not.
when it succeeds it will put the int value in your variable (minutes in this case) and return true.
if it can't parse to an integer value it will return false.

More advanced example:
```
int minutes;
int seconds;
if (!int.TryParse(minutes.Trim(), out minutes)
  || !int.TryParse(seconds.Trim(), out seconds)
{
   //failed to get the minutes or the seconds
   return Fail("Unable to get the wait time.");
   //or you can comment the above line and uncomment this line to use the random claim timings the user has entered in case the script cant get the time.
   //return GetWaitSetting();
}

//parsing worked - return the time we need to wait
return (minutes * 60) + seconds;
```

AnonymousDeveloper65535: Oh thanks!

AnonymousDeveloper65535:
> Thanks @AnonymousDeveloper65535 for taking the time on doing a PR with this script. As aviram pointed out - its an exact copy of freedoge.co.in - so its a bit suspicious if it turns out to be scam or not.

I will keep using the site to see if i can get a withdrawl and will show proof if I can get to it

freedogeon.com/FreeDogeonScript
```diff
@@ -84,16 +84,23 @@ public override FaucetSettings Settings
         {
             return 0;
         }
-
-        int min = Convert.ToInt32(minutes.Trim());
-        int sec = Convert.ToInt32(seconds.Trim());
+if (!int.TryParse(minutes.Trim(), out minutes))
+{
+  //unable to parse.
+  return base.GetFaucetWaitTime();
```
faucetcollector: My bad for providing the wrong example:
`base.GetFaucetWaitTime();` will always `return 0;` - which results in retrying directly.
better use `return GetWaitSetting();`
that will use the random wait timings the user has setup on the script.

AnonymousDeveloper65535:
its legit here is the proof of transaction: https://tinyurl.com/vs3ldux

faucetcollector:
Cool, I will merge your PR.

PR #286, opened 2020-03-26
Update MoonBitco.in.fc

cgerabit:
Fix script to prevent same captcha sended multiple times

### DIFF
--- a/MoonBitco.in/MoonBitco.in.fc
+++ b/MoonBitco.in/MoonBitco.in.fc
@@ -18,8 +18,8 @@ public class MoonBitcoinScript : FaucetScript
                 //We have a setting to get the email address used for the coinpot account
                 new FaucetSetting() { Name = "Email", Display = "Coinpot email addres", Type = EditorType.TextBox, Required = true },
                 //Option to use SolveMedia or reCAPTCHA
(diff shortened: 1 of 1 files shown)


MoonBitco.in/MoonBitco.in.fc
```diff
@@ -20,6 +19,7 @@ public class MoonBitcoinScript : FaucetScript
                 new FaucetSetting() { Name = "Email", Display = "Coinpot email addres", Type = EditorType.TextBox, Required = true },
                 //Option to use SolveMedia or reCAPTCHA
                    new FaucetSetting(){Name="SwitchC",Display="SwitchCaptcha is the other is not visible",Type=EditorType.CheckBox,Default=true},
+                    new FaucetSetting(){Name="NotReload", Display="Not reload site on captcha error",Type=EditorType.CheckBox,Default=true},
```
faucetcollector: The message is a bit confusing - "not reload the site" doesn't tell the people what it does when unchecked. Its better to tell them about that it will do 3 attempts to solve SolveMedia.
Can you change the Display to: "Do three SolveMedia attempts when failed to answer". And maybe instead of NotReload name it "SolveMediaRetry" or something like that.

And can you fix the text on SwitchC to: "Switch Captcha if the other is not visible"

PR #290, opened 2020-03-27
Create ByBit.com.fc

aviramdayan:
create

### DIFF
--- a/ByBit.com/ByBit.com.fc
+++ b/ByBit.com/ByBit.com.fc
@@ -210,7 +210,8 @@ public class bybit : FaucetScript
 		
     }	//end ad
 	
-		
+        return GetWaitSetting();  
+
    //let Faucet Collector continue.
    return base.GetFaucetWaitTime();
  }
--- a/ByBit.com/ByBit.com.fc
+++ b/ByBit.com/ByBit.com.fc
@@ -209,8 +209,6 @@ public class bybit : FaucetScript
         return GetWaitSetting();  
 		
     }	//end ad
-	
-        return GetWaitSetting();  
 
    //let Faucet Collector continue.
    return base.GetFaucetWaitTime();
--- a/ByBit.com/ByBit.com.fc
+++ b/ByBit.com/ByBit.com.fc
@@ -271,7 +271,7 @@ public class bybit : FaucetScript
 		GoToUrl("https://www.bybit.com/en-US/prize_page");
 		
 	    Wait();
-	    Wait(4);
+	    Wait(2);
 
         var Button_Claim = ElementByXPath("//div[@class='prize__banner-info']//span[contains(text(),'Claim Now') ]");  
         if (IsVisible(Button_Claim))
@@ -279,15 +279,11 @@ public class bybit : FaucetScript
 		Click(Button_Claim);		
 		}
 		
-	    Wait();		
 	    Wait(2);
 		
 		Recaptcha_Function (); //call if apper	
 		
-	    Wait();		
 	    Wait(4);
-		
-	    Driver.Navigate().Refresh(); //Refresh		
 		
         var Success_Claim = ElementByXPath("//div/h3/b[contains(text(),'have won')] "); //("//div[@class='alert alert-success']")
         if (IsVisible(Success_Claim))
@@ -304,8 +300,6 @@ public class bybit : FaucetScript
         return base.DoSolveFaucet();
     }
 	
-	
-
 	
 	
     public override int AfterSolveFaucet()
--- a/ByBit.com/ByBit.com.fc
+++ b/ByBit.com/ByBit.com.fc
@@ -1,0 +1,360 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 27/03/2020 10:30AM - Version 1
+//Last Update: 27/03/2020 10:30AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+
+
+public class bybit : FaucetScript
+{
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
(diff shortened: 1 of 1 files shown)


ByBit.com/ByBit.com.fc
```diff
@@ -0,0 +1,360 @@
...
+	    Wait(4);
+		
+	    Driver.Navigate().Refresh(); //Refresh		
+		
+        var Success_Claim = ElementByXPath("//div/h3/b[contains(text(),'have won')] "); //("//div[@class='alert alert-success']")
+        if (IsVisible(Success_Claim))
+		{
+		Log("FaucetClaim - Success_Claim ");  
+        ad = true; //call to timer / and do success
+		}
```
cgerabit: if you refresh the site after the claim the success message will be deleted

ByBit.com/ByBit.com.fc
```diff
@@ -0,0 +1,360 @@
...
+            if (result > 0)
+            {
+                return Fail("Fail solving captcha");
+            }
+            RecaptchaUtility utility = new RecaptchaUtility(this);
+            utility.DoSolve();
+		  }
+
+		return 0;
+        }//end VOID_Template_Function
```
cgerabit: the script may return an exception in the log, since in the line 276 you only click the button if it's visible and the function continue in execution the RecaptchaFunction will be called although recaptcha is not visible, if there is not a captcha base.DoSolveCaptcha will return a value <=0 and utility.DoSolve will be executed, I think the script will not close, but the user will receive the error message.

ByBit.com/ByBit.com.fc
aviramdayan: remove refresh

ByBit.com/ByBit.com.fc
```diff
@@ -0,0 +1,354 @@
...
+        var Success_Claim = ElementByXPath("//div/h3/b[contains(text(),'have won')] "); //("//div[@class='alert alert-success']")
+        if (IsVisible(Success_Claim))
+		{
+		Log("FaucetClaim - Success_Claim ");  
+        ad = true; //call to timer / and do success
+		}
+		
+      } //end FaucetClaim
+
+      return GetWaitSetting();
```
cgerabit: the second run will possibly be skipped since you set ad = true but GetFaucetWaitTime will never run since it is returning GetWaitSetting before that happens (that means that ad never be setted as false) , so in the second run it will already have the value true and the time will be returned without claim

ByBit.com/ByBit.com.fc
aviramdayan: the recaptcha visible only if claim button exist

cgerabit: I am  seeing that you verified if the recaptcha iframe is visible so this probably will not happen, it is still advisable to cut the function if you cannot go to the claim page but I leave it to your liking

ByBit.com/ByBit.com.fc
aviramdayan: i work with this method in all my scripts, all return well.. to "ad" and to the end "second" return

cgerabit: well i think the script instance is used as long as you don't press the stop button, so this should ignore the second claim since ad will keep in true. but since I'm not sure I think that @faucetcollector can tell us better

aviramdayan: just try it

faucetcollector: Now the claim counter on the bottom of FC will never be incremented.
An cg is right, the `ad` will stay true. The `Start` method is only called once. Only when you stop/start again its compiling the script again and calling the `Start` method

If you just set the `SuccessXPath` to //div/h3/b[contains(text(),'have won')] then you can leave all this out.

aviramdayan: as i said we just stuck on nothing here, 
i do it in all my script in all until now all work like i wanted
i know he stay true, the "ad" focus on "Success", 
if not exist - so at the end he return to the second (backup), 
why to do story on nothing here?

cgerabit: Hey, i did a test script as you said before and tested it this are the results: 
`03-29 00:14:11: None: Starting faucets..
03-29 00:14:16: ExampleScript: Init.
03-29 00:14:17: ExampleScript: Starting ExampleScript
03-29 00:14:34: ExampleScript: In DosolveFaucet
03-29 00:14:34: ExampleScript: Retorning time in DoSolveFaucet
03-29 00:14:34: ExampleScript: Closing browser..
03-29 00:14:34: ExampleScript: Closing ExampleScript
03-29 00:14:46: ExampleScript: Waiting 1 minute before continuing.
03-29 00:15:46: ExampleScript: Starting ExampleScript
03-29 00:15:57: ExampleScript: Retorning time in get faucetwait time
03-29 00:15:57: ExampleScript: Closing browser..
03-29 00:15:57: ExampleScript: Closing ExampleScript
03-29 00:16:09: ExampleScript: Waiting 1 minute before continuing.`

as you can see first run is executed DoSolveFaucet and wait 1 minute with GetWaitSetting, in the second run it directly wait without execute DoSolveFaucet, this happens because the ad variable is still true, so since you are not using it at the moment it keep set true. then in the next execution the script will check if(ad==true) in GetFaucetWaitTime and how in your past execution your didn't setted it as false the method will return the wait time and DoSolveFaucet will not be executed

I uploaded the script to a new branch in case you want to see it yourself

https://github.com/Faucet-Collector/scripts/blob/feature/ExampleScript/ExampleScript/ExampleScript.fc

aviramdayan: i leave the 3 (last commit) here,  
you stuck & do story on nothing here (all return & working well).

ByBit.com/ByBit.com.fc
```diff
@@ -210,7 +210,8 @@ public class bybit : FaucetScript
 		
     }	//end ad
 	
-		
+        return GetWaitSetting();  
+
```
cgerabit: since there is not any conditional the script will return wait time 100% of the times

PR #300, opened 2020-04-01
Update CoinPay4u.fc

cgerabit:
fix recognition of advertisements already viewed

### DIFF
--- a/CoinPayU/CoinPay4u.fc
+++ b/CoinPayU/CoinPay4u.fc
@@ -232,7 +232,7 @@ class ScriptCoinPayU: FaucetScript
             Wait(1);
             int n = 0;
 			int PastAd=0;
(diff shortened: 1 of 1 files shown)


CoinPayU/CoinPay4u.fc
```diff
@@ -232,7 +232,7 @@ class ScriptCoinPayU: FaucetScript
             Wait(1);
             int n = 0;
 			int PastAd=0;
-            var Ads = ElementsByXPath("//div[@class='col-lg-9 ags-left no-padding-right col-lg-9 col-xl-9']//div[contains(@class,'col-12 text-overflow ags-description pull-left')]");
+            var Ads = ElementsByXPath("//div[@class='clearfix container-fluid bgwhite']//div[@class='col-lg-9 ags-left no-padding-right col-lg-9 col-xl-9']//div[contains(@class,'col-12 text-overflow ags-description pull-left')]");
```
faucetcollector, line 235: maybe using a `contains` would be better, if not too many similiar elements with same classes are there.

PR #302, opened 2020-04-01
Update FreeBitco.in.fc

cgerabit:
*Add option to claim largest bonus in every setting
*Added a separated list for every available settings
*Modified accessibility of some methods of RedemSetting class


FreeBitco.in/FreeBitco.in.fc
```diff
@@ -483,11 +548,12 @@ public class FreeBitcoinScript_ : FaucetScript
     {
 
 
+        var RewardPoints=this.RewardPoints;
```
faucetcollector, line 551: Dont think this one is needed

PR #308, opened 2020-04-05
Update doge-bitsfree.net.fc

cgerabit:
*Add close chat to avoid captcha interruption
*Add wait time before try to claim faucet and also added a message if faucet is not correctly loaded

### DIFF
--- a/doge-bitsfree.net/doge-bitsfree.net.fc
+++ b/doge-bitsfree.net/doge-bitsfree.net.fc
@@ -342,8 +342,8 @@ public class Dogebitsfree: FaucetScript
             LoadingFaucet=ElementByXPath ( "//div[@id='loadingFaucet']" );
             if( DateTime.Now<timeout )
             {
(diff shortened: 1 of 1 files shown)


doge-bitsfree.net/doge-bitsfree.net.fc
```diff
@@ -323,15 +331,20 @@ public class Dogebitsfree: FaucetScript
...
         var LoadingFaucet=ElementByXPath("//div[@id='loadingFaucet']");
         DateTime timeout=DateTime.Now.AddSeconds(20);
-        while ( IsVisible ( LoadingFaucet ) && DateTime.Now<timeout )
+        while ( IsVisible ( LoadingFaucet ))
         {
             Wait ( );
             LoadingFaucet=ElementByXPath ( "//div[@id='loadingFaucet']" );
+            if( DateTime.Now<timeout )
+            {
+                Log ( "Seems that there is a problem and faucet is not loading" );
```
faucetcollector, line 345: It should `return` after this Log

PR #311, opened 2020-04-05
Create Paradice.in.fc

aviramdayan:
create

### DIFF
--- a/Paradice.in/Paradice.in.fc
+++ b/Paradice.in/Paradice.in.fc
@@ -1,0 +1,402 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 06/04/2020 00:15PM - Version 1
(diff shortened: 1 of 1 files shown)


Paradice.in/Paradice.in.fc
```diff
@@ -0,0 +1,402 @@
...
+      if (GetBoolSetting("FaucetClaim"))
+      {
+	
+        //else go to normal login
+		GoToUrl("https://paradice.in/account?modal=faucet");
+		
+	    Wait();
+	    Wait(4);
+
+        var WaitTimerText = ElementByXPath("//div[contains(@class,'faucet-limit')]");
```
faucetcollector, line 291: duplicate code from above in `GetFaucetWaitTime` and best to use int.TryParse or add a try catch, as when it wont be able to get the timer text then it will throw an exception and the script will restart.

PR #326, opened 2020-04-11
Update my.dropz.xyz.fc

aviramdayan:
update-fix-ViewAds-element-final


aviramdayan:
send new commit - general improvement

PR #332, opened 2020-04-12
Claimbits

avril-verhaeghen:


faucetcollector:
Maybe we can replace the existing one with this one instead of creating a new script for the same.
https://faucetcollector.com/faucets/details/f7f90193129a450d80801fb1c80d2eae

That is this one:
https://github.com/Faucet-Collector/scripts/tree/master/Claimbits%20Shortlinks

PR #339, opened 2020-04-17
Create CryptoMiningGame.fc

cgerabit:
*Initial commit for the script that will claim in the V2 of the site:
Available Options:
Beat antibot
Automatic Activate mining
Automatic Jobs

### DIFF
--- a/CryptoMiningGame/CryptoMiningGame.fc
+++ b/CryptoMiningGame/CryptoMiningGame.fc
@@ -1,0 +1,510 @@
+﻿using FaucetCollector.Script;
+using OpenQA.Selenium;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+//css_inc Shortlinkutilities
+//css_ref HumanActions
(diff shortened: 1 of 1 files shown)


CryptoMiningGame/CryptoMiningGame.fc
```diff
@@ -0,0 +1,510 @@
...
+    {
+        return ElementByXPath ( "//div[@id='antispyware_block']" )!=null;
+    }
+
+    public bool SolveAntibot ( )
+    {
+        HumanActions.DriverHumanActions driver=new HumanActions.DriverHumanActions(Driver);
+        var Timer =ElementByXPath("//span[@id='beattime']");
+        var Block=ElementByXPath("//div[@id='antispyware_block']");
+        while ( IsVisible ( Timer )&&IsVisible ( Block ) )
```
faucetcollector, line 325: Best to add some protection to exit the while loop after a while in case something is wrong

PR #400, opened 2020-05-16
Create Payoneer.com.fc

aviramdayan:
create


faucetcollector:
Payoneer doesn't have a faucet or anything they can earn.

PR #404, opened 2020-05-18
Update Adbit.pro.fc

aviramdayan:
fix-xpath

### DIFF
--- a/Adbit.pro/Adbit.pro.fc
+++ b/Adbit.pro/Adbit.pro.fc
@@ -2,7 +2,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 10/04/2020 17:30PM - Version 1
-//Last Update: 18/06/2020 02:40PM - Version 2
+//Last Update: 18/05/2020 17:10PM - Version 3
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


Adbit.pro/Adbit.pro.fc
```diff
@@ -277,36 +277,52 @@ public override int DoSolveFaucet()
...
+
+			if (IsVisible(WaitSec))
 		    {
+              Log("SurfAds - wait");
+              Wait();			  
+			}
+
+            if (IsVisible(ElementByXPath ("//div[contains(@id,'spinner') and contains(@style,'none') ] ") )) 
+            {	
+              //CheckForPopupsEnabled = true; //enable
```
faucetcollector, line 319: are you keeping this on false on purpose?
it will keep all popups open

aviramdayan: yes, this is the only way according to this site and xpath.. to keep..
take a look at xpath in sitw all good, it should return to this:
`
if (IsVisible(ElementByXPath ("//div[contains(@id,'spinner') and contains(@style,'none') ] ") )) 
`
but he doesn't do that..

PR #405, opened 2020-05-19
Add code filter for all moonfaucets

cgerabit:
*Add code filter for all moonfaucets


JReming85:
sweet, I also found these will allow coinpot to work without blocking claims
image

PR #425, opened 2020-06-07
Create HauGames.com.fc

aviramdayan:
create

### DIFF
--- a/HauGames.com/HauGames.com.fc
+++ b/HauGames.com/HauGames.com.fc
@@ -1,0 +1,411 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 01/02/2020 10:30AM - Version 1
+//Last Update: 01/02/2020 10:30AM - Version 1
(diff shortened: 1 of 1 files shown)


HauGames.com/HauGames.com.fc
```diff
@@ -0,0 +1,411 @@
...
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+using System.Drawing;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+//css_inc ShortLinkUtilities
+//css_inc RecaptchaUtilities
+//css_inc AntiBotLinksUtility
```
faucetcollector, line 19: best is to only include what you are using

aviramdayan: this "using" i sometimes forget, 
so it's better for me that he stay in the top :)

PR #429, opened 2020-06-09
Create Claimbtc.io.fc

aviramdayan:
create

### DIFF
--- a/Claimbtc.io/Claimbtc.io.fc
+++ b/Claimbtc.io/Claimbtc.io.fc
@@ -1,0 +1,409 @@
+
+//==================//
+//script Written by: ••AviramDayan•• 01/02/2020 10:30AM - Version 1
+//Last Update: 09/06/2020 21:10PM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+using System.Drawing;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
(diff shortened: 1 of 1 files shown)


Claimbtc.io/Claimbtc.io.fc
```diff
@@ -0,0 +1,409 @@
...
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+using System.Drawing;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+//css_inc ShortLinkUtilities
```
faucetcollector, line 16: Please next time only include those that you actually use. In case we ever delete one we now have to update all scripts. Also this slow things down as it will also compile these files in the script.

Claimbtc.io/Claimbtc.io.fc
```diff
@@ -0,0 +1,409 @@
...
+        public int Time_Function()
+        {
+			
+           var TimeVAR = ElementByXPath("//button[@class='claim' and @disabled] ");
+
+           if (IsVisible(TimeVAR))
+           {
+	         Log("ClaimBTC - WaitTimerText apper - return to site time"); 
+             int min,sec;
+             min = Convert.ToInt32(TimeVAR.Text.Trim().Replace(" ", "").Split(':')[0])*60;
```
faucetcollector, line 367: ```int.TryParse``` or use ```try { .. } catch { return GetWaitSetting() }```

PR #451, opened 2020-06-20
Update TrustDice.Win.fc

aviramdayan:
fix-and-order

### DIFF
--- a/TrustDice.Win/TrustDice.Win.fc
+++ b/TrustDice.Win/TrustDice.Win.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 13/11/2019 09:00AM - Version 1
-//Last Update: 24/04/2020 04:00PM - Version 4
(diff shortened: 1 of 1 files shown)


TrustDice.Win/TrustDice.Win.fc
```diff
@@ -601,11 +560,49 @@ public class TrustDice : FaucetScript
...
+  return 0;  
   }// end SolveCaptcha_Function
   //=========//
 
 
+	   	  
+ //==============//
+ //Time_Function
+ //=============//
+ public int Time_Function()
```
faucetcollector, line 572: Better add a try catch in here.. 
Getting the elements has not optimal xpath and you don't check if it did find the element.
So it will get a NullReferenceException if it cannot find it, or throw another exception if it doesnt contain an int.

aviramdayan: You're right dear, but right now i don't have experience with try/catch - so slowly in the future.

PR #452, opened 2020-06-20
Create SimpleBits.io.fc

aviramdayan:
SimpleBits.io/create

### DIFF
--- a/SimpleBits.io/SimpleBits.io.fc
+++ b/SimpleBits.io/SimpleBits.io.fc
@@ -1,0 +1,430 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 01/02/2020 10:30AM - Version 1
+//Last Update: 21/06/2020 00:50AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Interactions;
+using System.Drawing;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+//css_inc ShortLinkUtilities
+//css_ref FaucetCollector.Utilities
+//css_inc RecaptchaUtilities
+//css_inc AntiBotLinksUtility
+//css_ref ImageProcessing
+//css_ref ProcessingImagesUsingLockBits
+//css_inc GetestCaptcha
+
+
+public class SimpleBits : FaucetScript
+{
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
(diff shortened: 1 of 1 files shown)


SimpleBits.io/SimpleBits.io.fc
```diff
@@ -0,0 +1,430 @@
...
+    {
+        get
+        {
+            return new FaucetSettings("https://SimpleBits.io")
+            {
+                new FaucetSetting(){Name="Email",Display="UserName only",Type=EditorType.TextBox,Required=true},
+                new FaucetSetting(){Name="Password",Display="Password",Type=EditorType.Password,Required=true},
+				
+                new FaucetSetting(){Name="ShortLinks",Display="ShortLinks",Type=EditorType.CheckBox,Default=true},		
+                new FaucetSetting(){Name="CaptchaLogin",Display="CaptchaLogin - ON - Recaptcha | OFF - SolveMedia ",Type=EditorType.CheckBox,Default=false},
```
faucetcollector, line 43: I think this might be confusing for customers.
"Use reCAPTCHA (Unchecked will use SolveMedia)"

SimpleBits.io/SimpleBits.io.fc
```diff
@@ -0,0 +1,430 @@
...
+        var VisitButton = ElementByXPath("//tr[contains(@draggable,'false')]//span[contains(text(),'Visit')]");
+        if (IsVisible(VisitButton))
+        {
+	     Wait(3);
+         Title = "";
+         Log("ShortLinks- VisitButton");  
+         CheckForPopupsEnabled = false; //disable	
+	     Click(VisitButton);
+	     Wait(2);
+		 Driver.Close();
```
faucetcollector, line 315: Does it always do a popup? Why can't the default popup checker not handle it?

aviramdayan: not always, only once, that is why I chose close
also it's different title and different workflow in this specific site,
in general you click on some link after it bring you and need do things in [1] and after back to [0] and check and close [1] popup

here when you click first he open in [1] window
after need to close the [0] (preview) and all stuff need to be here now

SimpleBits.io/SimpleBits.io.fc
aviramdayan: is the same, i do it in all my scripts, but i'll change

SimpleBits.io/SimpleBits.io.fc
faucetcollector: In case there is no popup this will fail and close the main window

PR #462, opened 2020-06-30
Update DiscordTip (Tip.cc).fc

aviramdayan:
update-some-elements-and-add-GridCoin-faucet

### DIFF
--- a/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
+++ b/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 05/12/2018 06:00AM 
-//Last Update: 30/06/2020 02:30AM - Version 59
+//Last Update: 01/07/2020 02:30AM - Version 59
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
```diff
@@ -988,6 +1002,33 @@ public class discordapp_ : FaucetScript
 
 
 
+   //=================//
+  //inviteServer_Elements_Function
+  //=================//
+  public int inviteServer_Elements_Function()
+  {
+	    var inviteServer1Button = ElementByXPath("//button[@type='button']/div"); //[text()='Accept Invite']
+	    var ContinueToDiscordServer1Button = ElementByXPath("//button[@class='marginTop40-i-78cZ button-3k0cO7 button-38aScr lookFilled-1Gx00P colorBrand-3pXr91 sizeLarge-1vSeWK fullWidth-1orjjo grow-q77ONN']/div[text()='Continue to Discord']");
```
faucetcollector, line 1011: sounds like a `contains(@class, 'some classes here')` might be a good option instead of this full exact match

aviramdayan: updated sent new commit

PR #486, opened 2020-07-15
Create FocusGames.io.fc

aviramdayan:
create

### DIFF
--- a/FocusGames.io/FocusGames.io.fc
+++ b/FocusGames.io/FocusGames.io.fc
@@ -1,0 +1,443 @@
+
+
+//==================//
+//script Written by: ••AviramDayan•• 15/07/2020 17:00AM - Version 1
+//Last Update: 15/07/2020 17:00AM - Version 1
+//==================//
+
+using System;
+using System.IO;
+using FaucetCollector.Script;
+using OpenQA.Selenium;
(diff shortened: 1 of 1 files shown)


FocusGames.io/FocusGames.io.fc
```diff
@@ -0,0 +1,443 @@
...
+    /// <summary>
+    /// List of Settings that will be shown in the bot when selecting this Faucet in the bot.
+    /// You can get the value the user entered with the methods: GetSetting("[Name of the FaucetSetting]"), GetBoolSetting and GetDateTimeSetting
+    /// You can also create a new setting value using SetSetting("[Name you want to use]", "value")
+    /// </summary>
+    public override FaucetSettings Settings
+    {
+        get
+        {
+            return new FaucetSettings("https://FocusGames.io")
```
faucetcollector, line 37: its focusgame the old faucetgame?

dunno if you can reuse games from the existing one:
https://github.com/Faucet-Collector/scripts/blob/master/FaucetGame%20-%20do%20not%20update/FaucetGame.fc

aviramdayan: amm maybe it's the same owner i don't know,
but it seems to me the system is a different ..

PR #495, opened 2020-07-17
Update BestChange.com.fc

aviramdayan:
fix-faucet-cap

### DIFF
--- a/BestChange.com/BestChange.com.fc
+++ b/BestChange.com/BestChange.com.fc
@@ -2,7 +2,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 13/09/2019 22:09PM - Version 1
-//Last Update: 11/01/2020 15:40PM - Version 7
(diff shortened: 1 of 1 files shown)


BestChange.com/BestChange.com.fc
```diff
@@ -281,13 +268,12 @@ public class firebitcoinfaucettech : FaucetScript
...
-		
-		
+
         var ClaimButton = ElementByXPath("//input[contains(@onclick,'get_bonus')]");
 		
 	    if (!IsVisible(ClaimButton)) //not visible
 	    { 
-	      Time_Function();//call to time options
+	      return Time_Function();//call to time options
 		  Log("ClaimButton - not visible - return to time");
```
faucetcollector, line 277: this log will never be shown, as it returns on the line above

aviramdayan: yep you're right, not bad in the next update.

PR #635, opened 2020-10-04
Update DiscordTip (Tip.cc).fc

aviramdayan:
add-2-more-airdrops-from-TipCC-and-update-the-withdrawal

### DIFF
--- a/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
+++ b/DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 05/12/2018 06:00AM 
-//Last Update: 24/09/2020 00:35AM - Version 62
+//Last Update: 04/10/2020 11:50AM - Version 63
 //==================//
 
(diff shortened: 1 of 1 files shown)


DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
```diff
@@ -932,6 +1068,41 @@ public class DiscordScript: FaucetScript
...
+  //Withdraw_Start_Function
+  //=================//
+  private int Withdraw_Start_Function()
+  {
+       Title = "Discord";	
+       GoToUrl("https://discordapp.com/activity");
+       Wait();
+       Driver.Navigate().Refresh();
+       Wait();
+       Driver.Navigate().Refresh();
```
faucetcollector, line 1082: why does it need so many reloads?

faucetcollector (approved):

DiscordTip (Tip.cc)/DiscordTip (Tip.cc).fc
aviramdayan: Sometimes after one load, there is a problem with search bar..

PR #648, opened 2020-10-06
Rename Coinfaucet/Coinfaucets - All.fc to Coinfaucets - All/Coinfauce…

aviramdayan:
…ts - All.fc

change the sub folder


faucetcollector:
What is the reason for this rename?
To be honest, I am not sure if renaming the folder is supported in the webhooks.. Its been a while since I've written those.

aviramdayan:
> 
> 
> What is the reason for this rename?
> To be honest, I am not sure if renaming the folder is supported in the webhooks.. Its been a while since I've written those.
just for organization, If it's getting us in trouble, then we'll give up.

PR #692, opened 2020-10-31
Feature/auto faucet manual

cgerabit:


cgerabit:
Problem with branch, seems that i've pushed also freetheter lol

PR #694, opened 2020-10-31
Update Stake.Com.fc

aviramdayan:
add-Reload-SiteTimeDate

### DIFF
--- a/Stake.Com/Stake.Com.fc
+++ b/Stake.Com/Stake.Com.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 26/05/2019 04:00PM
-//Last Update: 31/10/2020 23:10AM - Version 23
+//Last Update: 01/11/2020 04:10AM - Version 23
 //==================//
 
 using System;
@@ -560,9 +560,10 @@ public class stake : FaucetScript
 	         Wait(5);	
 			   
              //site time with date - example 10:00 PM | 10:00 AM
-             var XpathTimeVAR = ElementByXPath("//div[contains(@class,'ModalFormWrap')]/p[1]/span/span");
-             if (IsVisible(XpathTimeVAR))
+             var CheckAvailableGeneral = ElementByXPath("//span[contains(text(),'Next claim available')] | //span[contains(text(),'None yet')]"); 
+             if (IsVisible(CheckAvailableGeneral))
              {
+               var XpathTimeVAR = ElementByXPath("//div[contains(@class,'ModalFormWrap')]/p[1]/span/span");
 		       var TimeVAR = DateTime.ParseExact(XpathTimeVAR.Text, "t", new CultureInfo("en-US") ); //t = Short time pattern - for all info here: predefined formats: https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings  | custom formats: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings
                Driver.Close();  // Close specific chrome window
                Driver.Quit();  // also Close the console app, that was used to kick off the chrome window
--- a/Stake.Com/Stake.Com.fc
+++ b/Stake.Com/Stake.Com.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 26/05/2019 04:00PM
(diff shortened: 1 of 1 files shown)


Stake.Com/Stake.Com.fc
```diff
@@ -40,7 +41,9 @@ public class stake : FaucetScript
                 new FaucetSetting() { Name = "ExclusiveBonusDrop_Telegram", Display = "Exclusive Bonus Drop - ON/OFF - in Telegram - 'Stake.com - Play Smarter' - Public Channel", Type = EditorType.CheckBox, Default =true},	
 				
                 new FaucetSetting() { Name = "Reload", Display = "Reload - ON/OFF - claim unlimited time if available", Type = EditorType.CheckBox, Default =true},	
-                new FaucetSetting(){Name="RakeBack",Display="RakeBack - ON/OFF - from Bronze level",Type=EditorType.CheckBox,Default=false},
+                new FaucetSetting() { Name = "Reload_SiteTimeDate", Display = "Reload_SiteTimeDate - ON/OFF - return to SiteTimeDate", Type = EditorType.CheckBox, Default =true},
```
faucetcollector, line 44: If I would see this option in a script I would have no idea what it would mean
"Reload_SiteTimeDate - ON/OFF - return to SiteTimeDate"

Stake.Com/Stake.Com.fc
```diff
@@ -541,19 +544,42 @@ public class stake : FaucetScript
...
+	         Wait();
+	         Wait(5);	
+			   
+             //site time with date - example 10:00 PM | 10:00 AM
+             var CheckAvailableGeneral = ElementByXPath("//span[contains(text(),'Next claim available')] | //span[contains(text(),'None yet')]"); 
+             if (IsVisible(CheckAvailableGeneral))
+             {
+               var XpathTimeVAR = ElementByXPath("//div[contains(@class,'ModalFormWrap')]/p[1]/span/span");
+		       var TimeVAR = DateTime.ParseExact(XpathTimeVAR.Text, "t", new CultureInfo("en-US") ); //t = Short time pattern - for all info here: predefined formats: https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings  | custom formats: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings
+               Driver.Close();  // Close specific chrome window
```
faucetcollector, line 568: I don't think this was needed

Stake.Com/Stake.Com.fc
aviramdayan: It does not matter so much, most people understand that, when they are active on the site.

Stake.Com/Stake.Com.fc
aviramdayan: in this case the `Driver.` must to be here.
sometimes below 5 min sometimes  above 5 min.

faucetcollector: but if they use advanced limits for example then this will interfere with that

Stake.Com/Stake.Com.fc
faucetcollector: Can you change it to something like 
"Use the time to wait before next claim from the website"

Stake.Com/Stake.Com.fc
```diff
@@ -40,7 +41,9 @@ public class stake : FaucetScript
                 new FaucetSetting() { Name = "ExclusiveBonusDrop_Telegram", Display = "Exclusive Bonus Drop - ON/OFF - in Telegram - 'Stake.com - Play Smarter' - Public Channel", Type = EditorType.CheckBox, Default =true},	
 				
                 new FaucetSetting() { Name = "Reload", Display = "Reload - ON/OFF - claim unlimited time if available", Type = EditorType.CheckBox, Default =true},	
-                new FaucetSetting(){Name="RakeBack",Display="RakeBack - ON/OFF - from Bronze level",Type=EditorType.CheckBox,Default=false},
+                new FaucetSetting() { Name = "Reload_SiteTimeDate", Display = "Reload Time - return to the exact time that found on the site", Type = EditorType.CheckBox, Default =true},
```
faucetcollector, line 44: Good, thats more clear now

Stake.Com/Stake.Com.fc
faucetcollector: Can you remove these?

faucetcollector (approved):

PR #770, opened 2020-12-19
Update Stake.Com.fc

aviramdayan:
add-more-wait-before-return-to-short-time-pattern

### DIFF
--- a/Stake.Com/Stake.Com.fc
+++ b/Stake.Com/Stake.Com.fc
@@ -576,10 +576,10 @@ public class stake : FaucetScript
              GoToUrl("https://stake.com/?currency=btc&modal=vip&tab=reload");
 	         Log("Go to try - Reload to return to time");	
 	         Wait();
-	         Wait(10);	
 			   
              //site time with date - example 10:00 PM | 10:00 AM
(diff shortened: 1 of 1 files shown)


Stake.Com/Stake.Com.fc
```diff
@@ -576,7 +576,7 @@ public class stake : FaucetScript
              GoToUrl("https://stake.com/?currency=btc&modal=vip&tab=reload");
 	         Log("Go to try - Reload to return to time");	
 	         Wait();
-	         Wait(5);	
+	         Wait(10);
```
faucetcollector, line 579: You can also add the wait time to the `ElementByXPath `method.

Like this:
             `var CheckAvailableGeneral = ElementByXPath("//span[contains(text(),'Next claim available')] | //span[contains(text(),'None yet')]", 10); `

Then it will stop after 10s and if it isnt found, or when it finds it within time it doesn't have to wait 10s upfront, so its always faster.

faucetcollector (approved):

Stake.Com/Stake.Com.fc
aviramdayan: done

faucetcollector (approved):

PR #818, opened 2021-01-11
Update BitSler.com.fc

aviramdayan:
fix-elements

### DIFF
--- a/BitSler/BitSler.com.fc
+++ b/BitSler/BitSler.com.fc
@@ -1,14 +1,20 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 05/11/2018 17:00PM
-//Last update - 03/09/2020 16:45PM - version 13
+//Last update - 11/01/2021 06:30AM - version 14
 //==================//
 
 using System;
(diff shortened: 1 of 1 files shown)


BitSler/BitSler.com.fc
```diff
@@ -181,14 +187,29 @@ bool ad;
...
+     var MenuClose = ElementByXPath("//div[@class='na-burger']");
+	 if (IsVisible(MenuClose))
+	 {
+	 Log("Menu - closed - we open it");
+     Click(MenuClose);
+	 }
+
+	 if (IsVisible(MenuOpen))
+	 {
+	 Log("Menu - already - opened");
```
faucetcollector, line 200: not really something that a customer needs. maybe we should introduce a debug level soon, so we can specify Debug, Info, Error etc when doing a log.

faucetcollector (approved):

BitSler/BitSler.com.fc
aviramdayan: ok this just in case because i didn't use with null in general..

PR #831, opened 2021-01-19
Update BitSler.com.fc

aviramdayan:
fix-dice-game

### DIFF
--- a/BitSler/BitSler.com.fc
+++ b/BitSler/BitSler.com.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 05/11/2018 17:00PM
-//Last update - 11/01/2021 06:30AM - version 14
+//Last update - 19/01/2021 19:40PM - version 15
 //==================//
 
 using System;
@@ -11,6 +11,7 @@ using OpenQA.Selenium;
 using OpenQA.Selenium.Interactions;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 //css_ref FaucetCollector.Utilities
 //css_inc RecaptchaUtilities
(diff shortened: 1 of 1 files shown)


BitSler/BitSler.com.fc
```diff
@@ -1393,62 +1381,79 @@ var Dice_Losing_Increase_Input_Var = ElementByXPath("//input[@class='losing__eve
...
-             Wait();
-			 
-			 
-            var Counterbreak = ElementByXPath("//button[contains(text(),'Start autobet')]");
-	        if (IsVisible(Counterbreak))
-	        {
-            break; 		
-	        ad=true;
-            }	
+        while (IsVisible(ElementByXPath("//div[contains(@class,'col-12 col-sm-6 mt-1')]//input[contains(@type,'tel') and @disabled]") ))
```
faucetcollector, line 1421: this while loop can be an endless loop in case the element never goes invisible due to whatever reason

BitSler/BitSler.com.fc
```diff
@@ -1393,62 +1381,79 @@ var Dice_Losing_Increase_Input_Var = ElementByXPath("//input[@class='losing__eve
...
 
-      }//end while
+        var Counterbreak = ElementByXPath("//button[contains(text(),'Start autobet')]");
+	    if (Counterbreak !=null) //Visible
+	    {
+        Log("break");				
+        break; 				
+        }	
 
+      }//end while
```
faucetcollector, line 1442: this while loop can also go indefinatly in case the start autobet is never found

faucetcollector: this one is fixed

BitSler/BitSler.com.fc
faucetcollector: this one still needs endless loop "protection"

aviramdayan: deleted /fixed

faucetcollector (approved):

PR #903, opened 2021-02-15
Added log entry to Ad-Type scripts every time you view an ad

stuffi3000:
As the "pop-up xxx closed" message was removed from the log, this is meant to replace it, giving the user accurate information whenever the script is still doing its job, and what ad is currently being watched.

### DIFF
--- a/AdLTC/AdLtc.fc
+++ b/AdLTC/AdLtc.fc
@@ -235,6 +235,7 @@ public class AdLtcScript : FaucetScript
                     Click(ButtonAd , false);
                     Wait(1);
                     Driver.SwitchTo( ).Window(Driver.WindowHandles[0]);
+					Log("Viewing ad: " + ElementByXPath("/html/body/div[1]/div[2]/div[2]/h4/b").Text);
 
                     var Timer = ElementByXPath("//h5[@id='timer']");
                     DateTime Timeout = DateTime.Now.AddSeconds(135);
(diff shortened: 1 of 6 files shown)


AdLTC/AdLtc.fc
```diff
@@ -235,6 +235,7 @@ public class AdLtcScript : FaucetScript
                     Click(ButtonAd , false);
                     Wait(1);
                     Driver.SwitchTo( ).Window(Driver.WindowHandles[0]);
+					Log("Viewing ad: " + ElementByXPath("/html/body/div[1]/div[2]/div[2]/h4/b").Text);
```
faucetcollector, line 238: This is not a very robust xpath for finding the element. When the website its html changes just a little it will break the script.
Also `ElementByXPath()` might return null when not found - then the script will throw an exception, because you can't do `.Text` on null.

stuffi3000:
Yeah, I'm aware, but as it's just a bold h4 title on the page, it's not really a singular element fit for a relative XPath. It doesn't have an ID, neither a name, it's just a h4 title. What I could do is put it into an if visible loop, so it doesn't break the script when it doesn't exist

faucetcollector:
Yes, if you put it into a `if (IsVisible(..))` then I am fine with keeping the "bad" xpath. Probably there are ways to improve the xpath tho, as you can use a lot of ways to reach a specific element (matching based on attributes, text searches, parent/nearby elements etc).

stuffi3000:
Yeah, I'll try to make it more dynamic as soon as I go back to it. I first wanted to grave the title of the page that was popping-up, but due to uBlock Origin there title in the header was empty haha

Pasuk2005:
hello 1) the larvelfauset script does not pass DDoS protection by Cloudflare what can I do about it? 2) the simplebits script does not log in can you check and fix it?

stuffi3000:
@pasuk2005
Please ask this question in the Discord support, this it not the place to ask such question

aviramdayan:
Hello @stuffi3000 You can open again the PR, when you are ready to fix.

PR #919, opened 2021-02-27
Create Tower.Bet.fc

aviramdayan:
create

### DIFF
--- a/Tower.Bet/Tower.Bet.fc
+++ b/Tower.Bet/Tower.Bet.fc
@@ -593,4 +593,4 @@ public class TowerBet : FaucetScript
         //=================//	
 	
 	
-} //public class dogeto
+} //public class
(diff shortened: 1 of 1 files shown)


Tower.Bet/Tower.Bet.fc
```diff
@@ -0,0 +1,596 @@
...
+         Click(ConnectDuckDice);
+        }
+
+		Wait(2);
+
+        var HandlesVAR = Driver.WindowHandles;
+        foreach (var MyWindowVAR in HandlesVAR)
+        {
+			  
+          CheckForPopupsEnabled = false;
```
faucetcollector, line 162: no need to turn it back on?

faucetcollector (approved):

Tower.Bet/Tower.Bet.fc
aviramdayan: nope

PR #951, opened 2021-03-26
Update Graviex.Net.fc

aviramdayan:
update

### DIFF
--- a/Graviex.Net/Graviex.Net.fc
+++ b/Graviex.Net/Graviex.Net.fc
@@ -1,7 +1,7 @@
 
 //==================//
 //script Written by: ••AviramDayan•• 14/03/2021 05:00AM - Version 1
-//Last Update: 14/03/2021 05:00AM - Version 1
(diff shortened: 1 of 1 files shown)


Graviex.Net/Graviex.Net.fc
```diff
@@ -377,6 +378,41 @@ public class graviex : FaucetScript
...
+//=================//
+public int Get_Tag_Numbers_OR_Specific_Function()
+{
+	
+  //Get "Specific" Tag Number from all index (Length)
+  var All_Elements = ElementsByXPath("//a");
+  var Clicked_Exist_Element = All_Elements.FirstOrDefault
+  (MyAllVars => 
+  {
+  Log("Found all buttons: " + MyAllVars.Text);
```
faucetcollector, line 393: don't think you need to output this in the logs all //a elements until it found the first match.

faucetcollector (approved):

PR #1065, opened 2021-05-24
Update and rename Coinfaucet/Coinfaucets - All.fc to Coinfaucets - Al…

aviramdayan:
…l/Coinfaucets - All.fc

update


faucetcollector (changes requested):
please put it in the same folder as before

PR #1074, opened 2021-06-06
Rollercoin bot is working now?

Lontron:
I want to buy a faucet collector's jaw and I want to know if the rollercoin bot still works? Captcha is solved based on its api automatically? Or it is solved by faucet collector , without api


faucetcollector:
Hi @Lontron 
Rollercoin is supported, the geetest captchas are currently solved within our software, without any external APIs.
But the more difficult geetest captchas where you need to click three icons are currently work in progress to be solved either internally or via FCCS.

